PR #15, opened 2019-11-01
Add gas metering to api

sasurobert:
Config and gas values to API calls.


arwen/context/arwen.go
```diff
@@ -7,13 +7,22 @@ import (
...
 
 	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
 	"github.com/ElrondNetwork/go-ext-wasm/wasmer"
 )
 
+// gasCostFile defines a flag for the path to the preferences toml configuration file
+var gasCostFile = cli.StringFlag{
+	Name:  "gasCost",
+	Usage: "The gas cost configuration file to load",
+	Value: "../config/config.toml",
```
camilbancioiu: What does `../config/config.toml` resolve to? Will it really resolve to the `config.toml` in Arwen, or will it try to open a `config.toml` relative to the Node's executable?

arwen/context/arwen.go
```diff
@@ -34,6 +43,7 @@ type vmContext struct {
 	blockChainHook vmcommon.BlockchainHook
 	cryptoHook     vmcommon.CryptoHook
 	imports        *wasmer.Imports
+	instance       wasmer.Instance
```
camilbancioiu: Is this intended as a step towards reusing a Wasmer instance?

arwen/elrondapi/bigintOps.go
```diff
@@ -132,6 +132,9 @@ func bigIntgetArgument(context unsafe.Pointer, id int32, destination int32) {
 
 	value := hostContext.GetOne(destination)
 	value.Set(args[id])
+
+	gasToUse := hostContext.GasSchedule().BigIntAPICost.BigIntGetArgument
+	hostContext.UseGas(gasToUse)
```
camilbancioiu: `bigIntgetArgument()` should consume gas even if the received argument `id` is invalid.

arwen/context/arwen.go
sasurobert: yes.

arwen/elrondapi/bigintOps.go
sasurobert: done

arwen/context/arwen.go
sasurobert: changed.

arwen/context/arwen.go
```diff
@@ -7,13 +7,22 @@ import (
...
 	"github.com/ElrondNetwork/go-ext-wasm/wasmer"
 )
 
+// gasCostFile defines a flag for the path to the preferences toml configuration file
+var gasCostFile = cli.StringFlag{
+	Name:  "gasCost",
+	Usage: "The gas cost configuration file to load",
+	Value: "../config/config.toml",
+}
+
```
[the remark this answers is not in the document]
sasurobert: we will have a map of opcodes, which will be read into the structure at vm creation time.

config/gasSchedule_test.go
```diff
@@ -0,0 +1,57 @@
...
+	assert.Nil(t, err)
+
+	fmt.Printf("%+v\n", erdOp)
+
+	ethOp := &EthAPICost{}
+	err = mapstructure.Decode(gasMap, ethOp)
+	assert.Nil(t, err)
+
+	fmt.Printf("%+v\n", ethOp)
+}
```
camilbancioiu: Optionally add a test with a `gasMap` that contains a `0` field and assert that  `CreateGasConfig()` returns an error.

PR #19, opened 2019-11-06
Configurable gas costs per individual WASM opcodes

camilbancioiu:
* Defines struct `WASMOpcodeCost`, which stores gas cost for each of the 410 WASM opcodes
* Handles loading the `WASMOpcodeCost` struct from a `gasMap`
* Handles converting the `WASMOpcodeCost` to an array of 410 `uint32` elements, to pass to `go-ext-wasm` for creating a metered Wasmer instance


arwen/context/arwen.go
```diff
@@ -226,6 +227,14 @@ func (host *vmContext) RunSmartContractCall(input *vmcommon.ContractCallInput) (
 	return vmOutput, nil
 }
 
+func opcode_costs_uniform_value(value uint32) *[wasmer.OPCODE_COUNT]uint32 {
```
sasurobert: function not used

go.mod
```diff
@@ -5,7 +5,8 @@ go 1.13
 require (
 	github.com/ElrondNetwork/elrond-vm-common v0.0.9
 	github.com/ElrondNetwork/elrond-vm-util v0.0.5
-	github.com/ElrondNetwork/go-ext-wasm v0.0.6
+	github.com/ElrondNetwork/go-ext-wasm v0.0.7
+	github.com/ElrondNetwork/managed-big-int v0.0.3
```
sasurobert: why managed-big-int ? it is not used.

arwen/context/arwen.go
camilbancioiu: Removed.

go.mod
camilbancioiu: Vestigial. Removed.

PR #23, opened 2019-11-11
Erc20 rust test

andrei-marinica:


cmd/test/debugArwen.go
```diff
@@ -9,6 +9,20 @@ import (
 )
 
 func main() {
+	if len(os.Args) != 2 {
+		panic("One argument expected - the path to the json test.")
+	}
+	jsonTestPath := os.Args[1]
+
+	err := controller.RunSingleIeleTest(
```
camilbancioiu: Maybe rename `RunSingleIeleTest()` to the generic `RunSingleVMTest()`?

cmd/test/executorRun.go
```diff
@@ -42,12 +43,22 @@ func newArwenTestExecutor() *arwenTestExecutor {
...
 	if len(value) == 0 {
 		return "", nil
 	}
-	fullPath := filepath.Join(testPath, value)
+	var fullPath string
+	if value == "wrc20_arwen.wasm" && te.erc20Path != "" {
+		fullPath = te.erc20Path
+	} else {
+		fullPath = filepath.Join(testPath, value)
+	}
```
camilbancioiu: Why is `value == "wrc20_arwen.wasm"` necessary here? Can't it be treated like any other WASM binary?

cmd/test/executorRun.go
```diff
@@ -42,12 +43,22 @@ func newArwenTestExecutor() *arwenTestExecutor {
 	}
 }
 
+func (te *arwenTestExecutor) setErc20Path(erc20Path string) *arwenTestExecutor {
```
camilbancioiu: Why is there a setter dedicated to the ERC20 test?

cmd/test/executorRun.go
andrei-marinica: we have the same tests for 3 different binaries, it was a quick and dirty solution. I'll replace it with something more palatable

cmd/test/debugArwen.go
andrei-marinica: good point, I changed the one with many tests but forgot about this one

PR #29, opened 2019-11-14
EN-4822/wasmer-reusing-resources

camilbancioiu:
The `VMContext` now stores a `wasmer.ImportObject` created during the initialization of the VM. The `ImportObject` is then passed to Wasmer whenever a SmartContract is called.

### DIFF
--- a/arwen/context/arwen.go
+++ b/arwen/context/arwen.go
@@ -150,7 +150,7 @@ func (host *vmContext) RunSmartContractCreate(input *vmcommon.ContractCreateInpu
 		return host.createVMOutputInCaseOfError(vmcommon.ContractInvalid), nil
 	}
 
-	defer host.instance.Close()
+	defer host.instance.Clean()
 
 	idContext := arwen.AddHostContext(host)
 	host.instance.SetContextData(unsafe.Pointer(&idContext))
--- a/arwen/context/arwen.go
+++ b/arwen/context/arwen.go
@@ -36,7 +36,6 @@ type vmContext struct {
 	blockChainHook vmcommon.BlockchainHook
 	cryptoHook     vmcommon.CryptoHook
 	imports        *wasmer.Imports
-	importObject   wasmer.ImportObject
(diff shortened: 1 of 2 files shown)


arwen/context/arwen.go
```diff
@@ -149,8 +158,6 @@ func (host *vmContext) RunSmartContractCreate(input *vmcommon.ContractCreateInpu
 		result = convertedResult.Bytes()
 	}
 
-	gasLeft := input.GasProvided.Int64()
-
 	// take out contract creation gas
 	gasLeft = gasLeft - int64(len(input.ContractCode))
```
sasurobert: calculate correct gas left

arwen/context/arwen.go
```diff
@@ -187,13 +195,13 @@ func (host *vmContext) RunSmartContractCall(input *vmcommon.ContractCallInput) (
 
 	var err error
 	opcode_costs := host.GasSchedule().WASMOpcodeCost.ToOpcodeCostsArray()
-	host.instance, err = wasmer.NewMeteredInstanceWithImports(contract, host.imports, uint64(gasLeft), &opcode_costs)
+	host.instance, err = wasmer.NewMeteredInstanceWithImportObject(contract, &host.importObject, uint64(gasLeft), &opcode_costs)
```
sasurobert: make op_code cost to be instantiated only once at arwen creation.

arwen/context/arwen.go
camilbancioiu: Done.

arwen/context/arwen.go
camilbancioiu: Done.

go.mod
```diff
@@ -3,9 +3,10 @@ module github.com/ElrondNetwork/arwen-wasm-vm
 go 1.13
 
 require (
+	github.com/ElrondNetwork/elrond-go v1.0.38 // indirect
```
sasurobert: this should be not needed.

PR #37, opened 2019-11-22
added compilation gas cost

sasurobert:


arwen/context/arwen.go
```diff
@@ -190,14 +193,21 @@ func (host *vmContext) RunSmartContractCreate(input *vmcommon.ContractCreateInpu
 	return vmOutput, err
 }
 
-func (host *vmContext) computeInitialCreateCost(input *vmcommon.ContractCreateInput) uint64 {
-	storePerByteCost := host.GasSchedule().BaseOperationCost.StorePerByte
-	createCost := host.GasSchedule().ElrondAPICost.CreateContract
+func (host *vmContext) computeInitialCodeCost(
```
camilbancioiu: Rename to `deductInitialCodeCost`.

arwen/context/arwen.go
```diff
@@ -138,13 +138,16 @@ func (host *vmContext) RunSmartContractCreate(input *vmcommon.ContractCreateInpu
 	host.scAddress = address
 	host.addTxValueToSmartContract(input.CallValue, address)
 
-	initialCreateCost := host.computeInitialCreateCost(input)
-
-	if input.GasProvided < initialCreateCost {
+	host.vmInput.GasProvided, err = host.computeInitialCodeCost(
+		input.GasProvided,
+		input.ContractCode,
+		host.GasSchedule().ElrondAPICost.CreateContract,
```
camilbancioiu: In `CreateNewContract()` below, there is a comment saying "creation costs different for ethereum and elrond". But where is this distinction made?

arwen/context/arwen.go
```diff
@@ -804,6 +840,8 @@ func (host *vmContext) execute(input *vmcommon.ContractCallInput) error {
 	convertedResult := arwen.ConvertReturnValue(result)
 	host.Finish(convertedResult.Bytes())
 
+	totalGasConsumed = input.GasProvided - gasLeft - newInstance.GetPointsUsed()
```
camilbancioiu: Variable not used?

arwen/context/arwen.go
```diff
@@ -804,6 +840,8 @@ func (host *vmContext) execute(input *vmcommon.ContractCallInput) error {
 	convertedResult := arwen.ConvertReturnValue(result)
```
camilbancioiu: Potential gas leak: my malicious SmartContract could run very expensive code, then trigger a `signalError()`, which will set `host.returnCode` to `vmcommon.UserError`. The lines above (see `if host.returnCode != vmcommon.Ok {...}`) do not deduct gas if this happens.

arwen/context/errors.go
```diff
@@ -11,3 +11,5 @@ var ErrFunctionRunError = errors.New("function run error")
 var ErrFuncNotFound = errors.New("function not found")
 
 var ErrReturnCodeNotOk = errors.New("return not is not ok")
+
+var ErrNotEnoughGas = errors.New("not enough gas")
```
camilbancioiu: Is this needed? We have `vmcommon.OutOfGas`.

sasurobert: no. as that is an error code only. here it is the full error.

arwen/context/arwen.go
sasurobert: used in defer.

arwen/context/arwen.go
sasurobert: ethereumEI and ElrondEI.

arwen/context/arwen.go
sasurobert: done.

PR #39, opened 2019-11-22
Fixes for ethereum, plus fix some gas checks

andreibancioiu:


arwen/context/arwen.go
```diff
@@ -949,7 +953,7 @@ func (host *vmContext) ReturnData() [][]byte {
 }
 
 func (host *vmContext) ClearReturnData() {
-	host.returnData = make([][]byte, 1)
+	host.returnData = make([][]byte, 0)
```
andreibancioiu: This was wrong before :)

PR #59, opened 2020-01-06
EN-5396: Asynchronous calls between SmartContracts, non-reentrant

camilbancioiu:
Implement `asyncCall`, the API method a SC may use to call another SC, with automatic synchronous execution when both SCs are in the same Shard. 

Currently, `asyncCall` is non-reentrant, which means that there are two restrictions:
* the called SC may not call `asyncCall`; doing so will result in undefined behaviour
* the `callBack` method of the caller SC may not call `asyncCall`; doing so will result in undefined behaviour

Complete error handling will be implemented after refactoring Arwen in the future.

Please test on the `elrond-go` branch `async-call-integration`, which provides extra tests in `integrationTests/multiShard/smartContract/scCallingSC_test.go`. More tests pending.

TODO: proper error handling
TODO: reentrant asynchronous calls


arwen/context/asyncCall.go
```diff
@@ -0,0 +1,286 @@
...
+
+import (
+	"math/big"
+
+	arwen "github.com/ElrondNetwork/arwen-wasm-vm/arwen"
+	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
+	"github.com/ElrondNetwork/go-ext-wasm/wasmer"
+)
+
+func (host *vmContext) handleAsyncCallBreakpoint(result wasmer.Value, err error) (*vmcommon.VMOutput, error) {
```
andreibancioiu: Add a comment with the meaning of `err`. It seems to be the WASMER `err` returned by SC function execution?

arwen/context/asyncCall.go
```diff
@@ -0,0 +1,286 @@
...
+	// TODO pass error to the SC callback (append as argument)
+	// TODO consume remaining gas
+	if err != nil {
+		vmOutputWithError := createVMOutputInCaseOfBreakpointError(err)
+		return mergeTwoVMOutputs(intermediaryVMOutput, vmOutputWithError), nil
+	}
+
+	intermediaryVMOutput = mergeTwoVMOutputs(intermediaryVMOutput, destinationVMOutput)
+
+	callbackCallInput, err := host.createCallbackContractCallInput(destinationVMOutput)
```
andreibancioiu: Is is possible to extract logic to a function called `executeCallback()`?

arwen/context/asyncCall.go
```diff
@@ -0,0 +1,286 @@
...
+		// TODO detect same Shard call - this makes the empty calledSCCode an error
+		// if intraShard {
+		//   vmOutputWithError := createVMOutputInCaseOfBreakpointError(err)
+		//   return mergeTwoVMOutputs(intermediaryVMOutput, vmOutputWithError)
+		// }
+		return intermediaryVMOutput, nil
+	}
+
+	// Start calling the destination SC, synchronously.
+	destinationCallInput, err := host.createDestinationContractCallInput()
```
andreibancioiu: Can we extract perhaps to a function called `callDestinationSCSync()`?

arwen/context/asyncCall.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func (host *vmContext) createCallbackContractCallInput(destinationVMOutput *vmcommon.VMOutput) (*vmcommon.ContractCallInput, error) {
+	arguments := destinationVMOutput.ReturnData
+	gasLimit := destinationVMOutput.GasRemaining
+	function := "callBack"
+
+	// Calculate what length would the Data field have, were it of the
+	// form "callback@arg1@arg4...
+	dataLength := len(function) + 1
```
andreibancioiu: Can we extract to a function called `getInputGasUsage` or something like that? And also write a unit test for it?

arwen/context/asyncCall.go
```diff
@@ -0,0 +1,286 @@
+package context
```
andreibancioiu: Add unit tests for this?

arwen/context/asyncCall.go
```diff
@@ -0,0 +1,286 @@
...
+
+func mergeVMOutputs(sender *vmcommon.VMOutput, destination *vmcommon.VMOutput, callback *vmcommon.VMOutput) *vmcommon.VMOutput {
+	vmOutput := &vmcommon.VMOutput{}
+	vmOutput = mergeTwoVMOutputs(vmOutput, sender)
+	vmOutput = mergeTwoVMOutputs(vmOutput, destination)
+	vmOutput = mergeTwoVMOutputs(vmOutput, callback)
+	return vmOutput
+}
+
+func mergeTwoVMOutputs(leftVMOutput *vmcommon.VMOutput, rightVMOutput *vmcommon.VMOutput) *vmcommon.VMOutput {
```
andreibancioiu: Add unit test.

arwen/context/arwen.go
```diff
@@ -708,6 +738,13 @@ func (host *vmContext) WriteLog(addr []byte, topics [][]byte, data []byte) {
 	host.logs[strAdr] = currLogs
 }
 
+func (host *vmContext) SetAsyncCallInfo(dest []byte, value []byte, gasLimit uint64, data []byte) {
```
sasurobert: this could have its own component.

arwen/context/arwen.go
```diff
@@ -733,6 +770,8 @@ func (host *vmContext) Transfer(destination []byte, sender []byte, gasLimit uint
 	senderAcc.BalanceDelta = big.NewInt(0).Sub(senderAcc.BalanceDelta, value)
 	destAcc.BalanceDelta = big.NewInt(0).Add(destAcc.BalanceDelta, value)
 	destAcc.Data = append(destAcc.Data, input...)
+
+	// TODO set gasLimit or accumulate gasLimit?
```
sasurobert: gasLimit can be deleted here.

arwen/context/arwen.go
```diff
@@ -266,6 +278,19 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
...
+	// cost for asynchronous calls must be locked temporarily, to make sure the
+	// callBack has enough gas to be called and executed
+	mustLockMinGasForAsyncCallBack := host.vmInput.CallType == vmcommon.AsynchronousCall
+	if mustLockMinGasForAsyncCallBack {
+		// Only the Elrond API supports asynchronous calls
+		if input.GasProvided < host.GasSchedule().ElrondAPICost.AsyncCallStep {
+			return host.createVMOutputInCaseOfError(vmcommon.OutOfGas), nil
+		}
+
+		input.GasProvided -= 2 * host.GasSchedule().ElrondAPICost.AsyncCallStep
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/arwen.go
```diff
@@ -195,7 +208,6 @@ func (host *vmContext) doRunSmartContractCreate(input *vmcommon.ContractCreateIn
 	}
 
 	vmOutput := host.createVMOutput(result)
-
 	return vmOutput, err
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/arwen.go
```diff
@@ -177,6 +189,7 @@ func (host *vmContext) doRunSmartContractCreate(input *vmcommon.ContractCreateIn
 	host.instance.SetContextData(unsafe.Pointer(&idContext))
 
 	_, result, err := host.callInitFunction()
+
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/asyncCall.go
camilbancioiu: Work in progres..

arwen/context/asyncCall.go
camilbancioiu: Work in progress.

arwen/context/asyncCall.go
camilbancioiu: Extracted as separate function, but unit test is pending.

arwen/context/asyncCall.go
camilbancioiu: This section has been refactored.

arwen/context/asyncCall.go
camilbancioiu: This section has been refactored.

arwen/context/arwen.go
```diff
@@ -304,20 +329,39 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
...
 	result, err := function()
+
+	// If some gas has been locked before calling the function to reserve it for
+	// the asynchronous callback, it will now be returned
+	if mustLockMinGasForAsyncCallBack {
+		input.GasProvided += 2 * host.GasSchedule().ElrondAPICost.AsyncCallStep
+	}
+
+	var vmOutput *vmcommon.VMOutput
+	if host.reachedBreakpoint(err) {
```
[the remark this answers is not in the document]
camilbancioiu: Removed `err` argument. And `err != nil` is handled at the last moment, using `defer`.

arwen/context/arwen.go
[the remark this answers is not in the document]
camilbancioiu: After refactoring, a new structure `AsyncCallInfo` was created, so the method name makes more sense now.

arwen/context/arwen.go
```diff
@@ -304,20 +329,39 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
 	}
 
 	result, err := function()
+
+	// If some gas has been locked before calling the function to reserve it for
+	// the asynchronous callback, it will now be returned
+	if mustLockMinGasForAsyncCallBack {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/arwen.go
```diff
@@ -266,6 +278,19 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
 		return host.createVMOutputInCaseOfError(vmcommon.ContractInvalid), nil
 	}
 
+	// If this call is an asynchronous call, a minimum of 2x the cost of the gas
+	// cost for asynchronous calls must be locked temporarily, to make sure the
+	// callBack has enough gas to be called and executed
+	mustLockMinGasForAsyncCallBack := host.vmInput.CallType == vmcommon.AsynchronousCall
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/arwen.go
```diff
@@ -266,6 +278,19 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
...
+	if mustLockMinGasForAsyncCallBack {
+		// Only the Elrond API supports asynchronous calls
+		if input.GasProvided < host.GasSchedule().ElrondAPICost.AsyncCallStep {
```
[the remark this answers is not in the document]
camilbancioiu: Done (almost as recommended).

arwen/context/interface.go
```diff
@@ -10,3 +13,15 @@ type BigIntContainer interface {
 	GetThree(id1, id2, id3 int32) (*big.Int, *big.Int, *big.Int)
 	IsInterfaceNil() bool
 }
+
+// ArgumentsParser defines the functionality to parse transaction data into arguments and code for smart contracts
+type ArgumentsParser interface {
```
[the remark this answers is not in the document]
camilbancioiu: Due to Go idioms, this interface should remain in place.

arwen/context/arwen.go
```diff
@@ -304,20 +329,39 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
...
+	}
+
+	var vmOutput *vmcommon.VMOutput
+	if host.reachedBreakpoint(err) {
+		vmOutput, err = host.handleBreakpoint(result, err)
+	}
+
 	if err != nil {
-		strError, _ := wasmer.GetLastError()
+		if err == ErrUnhandledRuntimeBreakpoint {
```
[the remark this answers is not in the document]
camilbancioiu: This section has been refactored.

arwen/context/arwen.go
```diff
@@ -304,20 +329,39 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
...
+			return host.CreateVMOutputInCaseOfErrorWithMessage(vmcommon.ExecutionFailed, ErrUnhandledRuntimeBreakpoint.Error()), nil
+		}
 
+		strError, _ := wasmer.GetLastError()
```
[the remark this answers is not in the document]
camilbancioiu: This section has been refactored. Printing errors to stdout has been removed anyway.

arwen/context/breakpoints_test.go
```diff
@@ -0,0 +1,59 @@
...
+import (
+	"math/big"
+	"testing"
+
+	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestMergingTwoOutputAccounts(t *testing.T) {
+	left := &vmcommon.OutputAccount{
```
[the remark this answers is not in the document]
camilbancioiu: This file has been deleted.

arwen/elrondapi/elrondei.go
```diff
@@ -360,7 +366,51 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
 
 	invBytes := arwen.InverseBytes(value)
 
-	hostContext.Transfer(dest, send, hostContext.BoundGasLimit(gasLimit), big.NewInt(0).SetBytes(invBytes), data)
+	hostContext.Transfer(dest, send, 0, big.NewInt(0).SetBytes(invBytes), data)
```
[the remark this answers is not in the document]
camilbancioiu: To prevent the misuse of `transferValue` for calling SmartContracts, because developers should use `asyncCall`.

arwen/context/asyncCall.go
camilbancioiu: `err` has been removed for now, since the entire error handling needs redesign.

arwen/context/arwen.go
```diff
@@ -280,7 +305,7 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
 
 	if err != nil {
 		fmt.Println("arwen Error", err.Error())
-		return host.createVMOutputInCaseOfError(vmcommon.ContractInvalid), nil
+		return host.CreateVMOutputInCaseOfErrorWithMessage(vmcommon.ContractInvalid, err.Error()), nil
```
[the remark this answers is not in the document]
camilbancioiu: Merged and renamed.

arwen/context/arwen.go
```diff
@@ -304,20 +329,39 @@ func (host *vmContext) doRunSmartContractCall(input *vmcommon.ContractCallInput)
 	}
 
 	result, err := function()
+
+	// If some gas has been locked before calling the function to reserve it for
+	// the asynchronous callback, it will now be returned
+	if mustLockMinGasForAsyncCallBack {
+		input.GasProvided += 2 * host.GasSchedule().ElrondAPICost.AsyncCallStep
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/arwen.go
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/arwen.go
```diff
@@ -336,6 +380,14 @@ func (host *vmContext) isInitFunctionCalled() bool {
 func (host *vmContext) createVMOutputInCaseOfError(errCode vmcommon.ReturnCode) *vmcommon.VMOutput {
 	vmOutput := &vmcommon.VMOutput{GasRemaining: 0, GasRefund: big.NewInt(0)}
 	vmOutput.ReturnCode = errCode
+	vmOutput.ReturnMessage = string(errCode)
+	return vmOutput
+}
+
+func (host *vmContext) CreateVMOutputInCaseOfErrorWithMessage(errCode vmcommon.ReturnCode, errMessage string) *vmcommon.VMOutput {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/context/arwen.go
camilbancioiu: Created the struct `AsyncCallInfo`, part of the `runtimeContext`.

arwen/context/arwen.go
camilbancioiu: `asyncCall` requires `gasLimit`, in case the call cannot be executed synchronously.

arwen/contexts/runtime.go
```diff
@@ -66,17 +72,36 @@ func (context *runtimeContext) InitStateFromContractCallInput(input *vmcommon.Co
...
-		callFunction: context.callFunction,
-		readOnly:     context.readOnly,
+		vmInput:       context.vmInput,
+		scAddress:     context.scAddress,
+		callFunction:  context.callFunction,
+		readOnly:      context.readOnly,
+		asyncCallInfo: context.asyncCallInfo,
 	}
 
 	context.stateStack = append(context.stateStack, newState)
```
sasurobert: let's add a limit of 10 states at most. 10 to be configurable.

arwen/contexts/runtime.go
```diff
@@ -230,8 +242,21 @@ func (context *runtimeContext) GetInitFunction() wasmer.ExportedFunctionCallback
 	return init
 }
 
-func (context *runtimeContext) MemLoad(offset int32, length int32) ([]byte, error) {
-	memory := context.instanceContext.Memory()
+func (runtime *runtimeContext) SetAsyncCallInfo(dest []byte, value []byte, gasLimit uint64, data []byte) {
```
sasurobert: as AsyncCallInfo is public this setter could get the AsyncCallInfo as parameter.

arwen/elrondapi/elrondei.go
```diff
@@ -377,9 +383,56 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
 	metering.UseGas(gasToUse)
 
 	invBytes := arwen.InverseBytes(value)
-	invValue := big.NewInt(0).SetBytes(invBytes)
-	boundGasLimit := metering.BoundGasLimit(gasLimit)
-	output.Transfer(dest, send, boundGasLimit, invValue, data)
+
```
sasurobert: delete empty line.

arwen/elrondapi/elrondei.go
```diff
@@ -377,9 +383,56 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
...
+	output.Transfer(dest, send, 0, big.NewInt(0).SetBytes(invBytes), data)
+
+	return 0
+}
+
+//export asyncCall
+func asyncCall(context unsafe.Pointer, destOffset int32, valueOffset int32, dataOffset int32, length int32) int32 {
```
sasurobert: no need for return as this is a blocking function and it will get out from the wasmer context.

arwen/elrondapi/elrondei.go
```diff
@@ -377,9 +383,56 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
...
+	runtime := arwen.GetRuntimeContext(context)
+	metering := arwen.GetMeteringContext(context)
+	output := arwen.GetOutputContext(context)
+
+	send := runtime.GetSCAddress()
+	dest, err := runtime.MemLoad(destOffset, arwen.AddressLen)
+	if withFault(err, context) {
+		return 1
```
sasurobert: instead of return SetRuntimeBreakPointValue.

arwen/host/breakpoints.go
```diff
@@ -8,19 +8,18 @@ import (
 func (host *vmHost) handleBreakpoint(
 	breakpointValue arwen.BreakpointValue,
 	result wasmer.Value,
-	err error,
 ) error {
 
 	if breakpointValue == arwen.BreakpointAsyncCall {
-		// return host.handleAsyncCallBreakpoint(result, err)
+		return host.handleAsyncCallBreakpoint(result)
 	}
```
sasurobert: delete empty line

arwen/host/breakpoints.go
```diff
@@ -8,19 +8,18 @@ import (
...
 
 	if breakpointValue == arwen.BreakpointSignalError {
 		return nil
 	}
```
sasurobert: delete empty line

arwen/host/execution.go
```diff
@@ -95,14 +91,20 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
...
-	gasForExecution, err := metering.DeductInitialGasForExecution(input, contract)
+	vmInput := runtime.GetVMInput()
+
+	err = metering.DeductInitialGasForExecution(contract)
+	if err != nil {
+		returnCode = vmcommon.OutOfGas
+		return vmOutput
+	}
+
+	err = metering.DeductAndLockGasIfAsyncStep()
```
sasurobert: make the deductandLockGasIfAsyncStep inside the DeductInitialGasForExecution method.

arwen/host/execution.go
```diff
@@ -117,29 +119,39 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
 	}()
 
 	result, returnCode, err := host.callSCMethod()
+
+	metering.UnlockGasIfAsyncStep()
+
 	if err != nil {
 		return vmOutput
```
sasurobert: i think here returnCode should be set to something is err is not nil.

sasurobert: just to make sure returnCode will yield error

arwen/contexts/runtime.go
```diff
@@ -66,17 +72,36 @@ func (context *runtimeContext) InitStateFromContractCallInput(input *vmcommon.Co
...
+		readOnly:      context.readOnly,
+		asyncCallInfo: context.asyncCallInfo,
 	}
 
 	context.stateStack = append(context.stateStack, newState)
 }
 
-func (context *runtimeContext) PushInstance() {
-	context.instanceStack = append(context.instanceStack, context.instance)
+func (runtime *runtimeContext) PopState() error {
```
sasurobert: i think popState should not return in error. Push and Pop always work in syncron.

arwen/host/execution.go
```diff
@@ -117,29 +119,39 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
...
+func (host *vmHost) ExecuteOnDestContext(input *vmcommon.ContractCallInput) (*vmcommon.VMOutput, error) {
 	host.PushState()
 
 	host.InitState()
 
 	host.Runtime().InitStateFromContractCallInput(input)
 	err := host.execute(input)
 
+	var vmOutput *vmcommon.VMOutput
+	vmOutput = nil
```
sasurobert: declare vmOutput before error. No need to make vmOutput nil. return nil, err if err != nil.

arwen/host/execution.go
```diff
@@ -117,29 +119,39 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
...
-	vmOutput = output.GetVMOutput(result)
+	output.FinishValue(result)
+	vmOutput = output.GetVMOutput()
 
 	return vmOutput
 }
 
-func (host *vmHost) ExecuteOnDestContext(input *vmcommon.ContractCallInput) error {
+func (host *vmHost) ExecuteOnDestContext(input *vmcommon.ContractCallInput) (*vmcommon.VMOutput, error) {
 	host.PushState()
```
sasurobert: let's make host.PushState() defer host.PopState()

go.sum
```diff
@@ -1,19 +1,54 @@
 github.com/ElrondNetwork/big-int-util v0.0.5 h1:e/9kK++9ZH/SdIYqLSUPRFYrDZmDWDgff3/7SCydq5I=
 github.com/ElrondNetwork/big-int-util v0.0.5/go.mod h1:96viBvoTXLjZOhEvE0D+QnAwg1IJLPAK6GVHMbC7Aw4=
+github.com/ElrondNetwork/elrond-vm-common v0.0.0-20191203115206-691b00a6e76a h1:/hGeyqQxE0E92BkMDs/1E6vRqJMA4nVhwSFol1dNopk=
 github.com/ElrondNetwork/elrond-vm-common v0.0.0-20191203115206-691b00a6e76a/go.mod h1:ZakxPST/Wt8umnRtA9gobcy3Dw2bywxwkC54P5VhO9g=
+github.com/ElrondNetwork/elrond-vm-common v0.0.9 h1:Ff8vEJSKChRfmp+TVo7AgciRkMXjL4+TbNin6LQ7xKw=
```
sasurobert: go mod tidy. no need for this big go.sum

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
+package host
+
+import (
+	"math/big"
+
+	arwen "github.com/ElrondNetwork/arwen-wasm-vm/arwen"
```
sasurobert: no need to redefine arwen

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+	_, err = host.ExecuteOnDestContext(callbackCallInput)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (host *vmHost) canExecuteSynchronously() bool {
+	runtime := host.Runtime()
```
sasurobert: take destination address and verify if it is in current shard. We can make a blockchain hook for it. canExecuteSynchronously will do only that.

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	arguments, err := argParser.GetArguments()
+	if err != nil {
+		return nil, err
+	}
+
+	gasLimit := asyncCallInfo.GasLimit - host.Metering().GasSchedule().ElrondAPICost.AsyncCallStep
```
sasurobert: if you do an underflow this will result in maxuint64 gasLimit.

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, arwen.ErrNotEnoughGas
+	}
+
+	contractCallInput := &vmcommon.ContractCallInput{
+		VMInput: vmcommon.VMInput{
+			CallerAddr:  sender,
+			Arguments:   arguments,
+			CallValue:   big.NewInt(0).SetBytes(asyncCallInfo.ValueBytes),
+			CallType:    vmcommon.AsynchronousCall,
+			GasPrice:    0,
```
sasurobert: there is probably no need for gasPrice in VMInput. but as it is there now, you may copy it from previous context

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+
+	arguments := destinationVMOutput.ReturnData
+	gasLimit := destinationVMOutput.GasRemaining
+	function := "callBack"
+
+	dataLength := host.computeDataLengthFromArguments(function, arguments)
+
+	gasToUse := metering.GasSchedule().ElrondAPICost.AsyncCallStep
+	gasToUse += metering.GasSchedule().BaseOperationCost.DataCopyPerByte * uint64(dataLength)
+	gasLimit -= gasToUse
```
sasurobert: for uint64 you have to check before if gasLimit is bigger than gasToUse.

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+	dest := runtime.GetSCAddress()
+
+	// Return to the sender SC, calling its callback() method.
+	contractCallInput := &vmcommon.ContractCallInput{
+		VMInput: vmcommon.VMInput{
+			CallerAddr:  sender,
+			Arguments:   arguments,
+			CallValue:   big.NewInt(0),
+			CallType:    vmcommon.AsynchronousCallBack,
+			GasPrice:    0,
```
sasurobert: copy gasPrice from previous context.

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+		Function:      function,
+	}
+
+	return contractCallInput, nil
+}
+
+func (host *vmHost) computeDataLengthFromArguments(function string, arguments [][]byte) int {
+	// Calculate what length would the Data field have, were it of the
+	// form "callback@arg1@arg4...
+	dataLength := len(function) + 1
```
sasurobert: dataLength = len(function) + SUM(len(argument)) + len(arguments)

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil
+	}
+
+	// Start calling the destination SC, synchronously.
+	destinationCallInput, err := host.createDestinationContractCallInput()
+	if err != nil {
+		return err
+	}
+
+	destinationVMOutput, err := host.ExecuteOnDestContext(destinationCallInput)
```
sasurobert: it is pretty easy to make these few calls in a recursive way. :)

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		return err
+	}
+
+	callbackCallInput, err := host.createCallbackContractCallInput(destinationVMOutput)
+	if err != nil {
+		return err
+	}
+
+	_, err = host.ExecuteOnDestContext(callbackCallInput)
```
sasurobert: is this last return not the final vmOutput which has to be merged with the rest ?

arwen/common.go
```diff
@@ -19,3 +19,10 @@ const (
 	BreakpointSignalExit  BreakpointValue = 4
 	BreakpointOutOfGas    BreakpointValue = 5
 )
+
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,84 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
 }
 
+func (context *meteringContext) DeductAndLockGasIfAsyncStep() error {
+	input := context.host.Runtime().GetVMInput()
+
+	if input.CallType == vmcommon.AsynchronousCall {
+		gasSchedule := context.GasSchedule().ElrondAPICost
+
+		input.GasProvided -= gasSchedule.AsyncCallStep
+		if input.GasProvided <= 0 {
```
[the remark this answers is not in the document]
camilbancioiu: This is bad. Fixed.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,84 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
+			return arwen.ErrNotEnoughGas
+		}
+
+		input.GasProvided -= gasSchedule.AsyncCallStep + gasSchedule.AsyncCallbackGasLock
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,84 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
 	}
 }
 
+func (context *meteringContext) DeductAndLockGasIfAsyncStep() error {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/output.go
```diff
@@ -177,8 +177,7 @@ func (context *outputContext) AddTxValueToAccount(address []byte, value *big.Int
 	destAcc.BalanceDelta = big.NewInt(0).Add(destAcc.BalanceDelta, value)
 }
 
-// adapt vm output and all saved data from sc run into VM Output
-func (context *outputContext) GetVMOutput(result wasmer.Value) *vmcommon.VMOutput {
+func (context *outputContext) GetVMOutput() *vmcommon.VMOutput {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/runtime.go
camilbancioiu: Indeed. Will come as a separate PR, because that change would require this method to return errors which need handling, and it would be messy in this PR. But the limit is required and will be added.

arwen/contexts/runtime.go
camilbancioiu: Changed.

arwen/elrondapi/elrondei.go
camilbancioiu: Done.

arwen/elrondapi/elrondei.go
camilbancioiu: Changed.

arwen/elrondapi/elrondei.go
camilbancioiu: Fixed: de-duplicated all definitions of `withFault`, which is now found in `helpers.go`, and `WithFault` always calls `runtime.SignalUserError()` which sets the breakpoint.

arwen/host/breakpoints.go
camilbancioiu: Done.

arwen/host/breakpoints.go
camilbancioiu: Done.

arwen/host/execution.go
camilbancioiu: Done.

arwen/host/execution.go
camilbancioiu: If `err != nil`, then `host.callSCMethod()` will have returned an appropriate value for `returnCode`. Then `err` is tested in the `defer` at the beginning of this method, along with the `returnCode`. Indeed, the code does not make this clear. Suggestions for rewriting?

arwen/contexts/runtime.go
camilbancioiu: True, `Pop` will only fail if there's a coding error in Arwen and the users cannot take any action for that case. Removed.

arwen/host/execution.go
camilbancioiu: Changed.

arwen/host/execution.go
camilbancioiu: Done.

go.sum
camilbancioiu: Done.

arwen/host/asyncCall.go
camilbancioiu: Done.

arwen/host/asyncCall.go
camilbancioiu: Added TODO for now, noted for next development steps.

arwen/host/asyncCall.go
camilbancioiu: Done.

arwen/host/asyncCall.go
camilbancioiu: Done.

arwen/host/asyncCall.go
camilbancioiu: Done.

arwen/host/asyncCall.go
camilbancioiu: Done.

arwen/host/asyncCall.go
camilbancioiu: Yes, async call reentrance will be much easier to implement after the last refactoring.

arwen/host/asyncCall.go
camilbancioiu: It is merged. `ExecuteOnDestContext()` calls `host.PopState()`, which calls `output.PopState()`, which merges the top two `VMOutput` entries in its stack. Afterwards, calling `output.GetVMOutput()` returns the merged outputs.

arwen/host/asyncCall.go
camilbancioiu: Right. In the current form, `computeDataLengthFromArguments()` ignores each empty argument. We need to change how `VMOutput.ReturnData` is handled, because empty entries are important, e.g. when interpreting `ReturnData` as arguments for `callBack()`. Currently, we skip these entries (see `Finish` functions) and this should be changed. Added a TODO and noted the issue.

arwen/contexts/metering.go
```diff
@@ -24,9 +25,10 @@ func NewMeteringContext(
 	}
 
 	context := &meteringContext{
-		gasSchedule:   gasCostConfig,
-		blockGasLimit: blockGasLimit,
-		host:          host,
+		gasSchedule:           gasCostConfig,
+		blockGasLimit:         blockGasLimit,
+		gasLockedForAsyncStep: false,
```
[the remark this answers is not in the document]
camilbancioiu: Changed to uint64.

arwen/contexts/bigint.go
```diff
@@ -32,18 +30,12 @@ func (context *bigIntContext) PushState() {
 	context.stateStack = append(context.stateStack, newState)
```
andreibancioiu: The new state has access to previous state's values / big integers, since `mapValues` isn't copied, but passed by reference. Is this correct & desired behavior?

arwen/crypto/cryptoei.go
```diff
@@ -37,7 +37,7 @@ func sha256(context unsafe.Pointer, dataOffset int32, length int32, resultOffset
 	crypto := arwen.GetCryptoContext(context)
 
 	data, err := runtime.MemLoad(dataOffset, length)
-	if withFault(err, context) {
+	if arwen.WithFault(err, context) {
```
sasurobert: there is no need for return for these functions if we are getting out of arwen execution with error and breakpoint. These returns are for the SC developer to know it was an error, so it might have some fallback mechanism implemented there.

arwen/elrondapi/elrondei.go
```diff
@@ -343,32 +349,32 @@ func blockHash(context unsafe.Pointer, nonce int64, resultOffset int32) int32 {
 	//TODO: change blockchain hook to treat actual nonce - not the offset.
```
sasurobert: TODO may be deleted as it was resolved.

arwen/elrondapi/elrondei.go
```diff
@@ -343,32 +349,32 @@ func blockHash(context unsafe.Pointer, nonce int64, resultOffset int32) int32 {
...
-func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, valueOffset int32, dataOffset int32, length int32) int32 {
+func transferValue(context unsafe.Pointer, destOffset int32, valueOffset int32, dataOffset int32, length int32) int32 {
 	runtime := arwen.GetRuntimeContext(context)
 	metering := arwen.GetMeteringContext(context)
 	output := arwen.GetOutputContext(context)
 
 	send := runtime.GetSCAddress()
 	dest, err := runtime.MemLoad(destOffset, arwen.AddressLen)
-	if withFault(err, context) {
+	if arwen.WithFault(err, context) {
```
sasurobert: there is no need to return anything from such a function if we get out from it with a breakpoint. so the decision has to be taken, whether to not return anything, and get out instantenausly from the wasmer context with error. Or leave the SC developer to do whatever he wants if a transfer call failed.

arwen/ethapi/ethereumei.go
```diff
@@ -487,15 +487,15 @@ func ethgetExternalCodeSize(context unsafe.Pointer, addressOffset int32) int32 {
...
+	if arwen.WithFault(err, context) {
 		return 0
 	}
 
 	gasToUse := metering.GasSchedule().EthAPICost.GetExternalCodeSize
 	metering.UseGas(gasToUse)
 
 	codeSize, err := blockchain.GetCodeSize(dest)
-	if withFault(err, context) {
+	if arwen.WithFault(err, context) {
```
sasurobert: as we have to keep the ethereum API, in places where we return 0 / 1 / -1 depending on failure or not, there is no need to set breakPoint and get out of arwen.

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,156 @@
...
+			GasProvided: gasLimit,
+		},
+		RecipientAddr: dest,
+		Function:      function,
+	}
+
+	return contractCallInput, nil
+}
+
+func (host *vmHost) computeDataLengthFromArguments(function string, arguments [][]byte) int {
```
andreibancioiu: Add unit test for this.

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,156 @@
...
+	metering := host.Metering()
+	runtime := host.Runtime()
+
+	arguments := destinationVMOutput.ReturnData
+	gasLimit := destinationVMOutput.GasRemaining
+	function := "callBack"
+
+	dataLength := host.computeDataLengthFromArguments(function, arguments)
+
+	gasToUse := metering.GasSchedule().ElrondAPICost.AsyncCallStep
```
andreibancioiu: Extract to separate function perhaps? And add unit test for it perhaps? `asyncCallInfo.GasLimit` will not get modified, but this is ok, right?

arwen/host/asyncCall.go
```diff
@@ -0,0 +1,156 @@
...
+			GasProvided: gasLimit,
+		},
+		RecipientAddr: asyncCallInfo.Destination,
+		Function:      function,
+	}
+
+	return contractCallInput, nil
+}
+
+func (host *vmHost) createCallbackContractCallInput(destinationVMOutput *vmcommon.VMOutput) (*vmcommon.ContractCallInput, error) {
```
andreibancioiu: Not sure, but this seems testable. Can we mock runtime perhaps and write a unit test for this function?

arwen/common.go
```diff
@@ -12,10 +12,18 @@ const (
 type BreakpointValue uint64
 
 const (
-	BreakpointNone        BreakpointValue = 0
-	BreakpointAbort       BreakpointValue = 1
-	BreakpointAsyncCall   BreakpointValue = 2
-	BreakpointSignalError BreakpointValue = 3
-	BreakpointSignalExit  BreakpointValue = 4
-	BreakpointOutOfGas    BreakpointValue = 5
+	BreakpointNone            BreakpointValue = 0
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,88 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
+
+		gasToConsume := gasSchedule.AsyncCallStep
+		gasToLock := gasSchedule.AsyncCallStep + gasSchedule.AsyncCallbackGasLock
+		if input.GasProvided <= gasToConsume+gasToLock {
+			return arwen.ErrNotEnoughGas
+		}
+		input.GasProvided -= gasToConsume + gasToLock
+
+		context.gasLockedForAsyncStep = gasToLock
+	} else {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,88 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
+		context.gasLockedForAsyncStep = 0
+	}
+
+	return nil
+}
+
+// UnlockGasIfAsyncStep will restore the previously locked gas, if the execution is an asynchronous call
+func (context *meteringContext) UnlockGasIfAsyncStep() {
+	if context.gasLockedForAsyncStep != 0 {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,88 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
 		contract,
 		0,
 		context.gasSchedule.BaseOperationCost.CompilePerByte,
 	)
-	return remainingGas, err
+	if err != nil {
+		return err
+	}
+
+	err = context.deductAndLockGasIfAsyncStep()
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/runtime.go
```diff
@@ -41,11 +44,13 @@ func NewRuntimeContext(
...
-	context.vmInput = &vmcommon.VMInput{}
-	context.scAddress = make([]byte, 0)
-	context.callFunction = ""
-	context.readOnly = false
+func (runtime *runtimeContext) InitState() {
+	runtime.vmInput = &vmcommon.VMInput{}
+	runtime.scAddress = make([]byte, 0)
+	runtime.callFunction = ""
+	runtime.readOnly = false
+	runtime.argParser, _ = vmcommon.NewAtArgumentParser()
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/elrondapi/elrondei.go
camilbancioiu: Done.

arwen/ethapi/ethereumei.go
camilbancioiu: Changed `WithFault` to be more flexible. Within the Ethereum API, no breakpoints will be set.

arwen/elrondapi/elrondei.go
camilbancioiu: Changed `WithError` to be more flexible. It may (or may not) set a breakpoint depending on the internal state of the Runtime context.

arwen/crypto/cryptoei.go
camilbancioiu: Changed `WithError` to be more flexible. Now the API may (or may not) set a breakpoint, depending on the internal state of the Runtime context. If execution is allowed to continue, the return value will be useful to the SC, as discussed.

arwen/contexts/bigint.go
```diff
@@ -32,18 +30,12 @@ func (context *bigIntContext) PushState() {
...
 }
 
-func (context *bigIntContext) PopState() error {
+func (context *bigIntContext) PopState() {
 	stateStackLen := len(context.stateStack)
-	if stateStackLen < 1 {
```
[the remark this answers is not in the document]
camilbancioiu: Only if Arwen code is written wrong, in which case the SC doesn't have any control over the crash, nor can it react. Both `PushState()` and `PopState()` are isolated from the API, for all contexts that have them, so the SC cannot cause errors with them. And if `asyncCall` handles the state stack properly, there should be no underflows.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,88 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
+// deductAndLockGasIfAsyncStep will deduct the gas for an async step and also lock gas for the callback, if the execution is an asynchronous call
+func (context *meteringContext) deductAndLockGasIfAsyncStep() error {
+	input := context.host.Runtime().GetVMInput()
+
+	if input.CallType == vmcommon.AsynchronousCall {
+		gasSchedule := context.GasSchedule().ElrondAPICost
+
+		gasToConsume := gasSchedule.AsyncCallStep
+		gasToLock := gasSchedule.AsyncCallStep + gasSchedule.AsyncCallbackGasLock
+		if input.GasProvided <= gasToConsume+gasToLock {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,88 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
 }
 
-func (context *meteringContext) DeductInitialGasForExecution(input *vmcommon.ContractCallInput, contract []byte) (uint64, error) {
-	remainingGas, err := context.deductInitialGas(
-		input.GasProvided,
+func (context *meteringContext) DeductInitialGasForExecution(contract []byte) error {
+	err := context.deductInitialGas(
 		contract,
 		0,
 		context.gasSchedule.BaseOperationCost.CompilePerByte,
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,88 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
+}
+
 func (context *meteringContext) BlockGasLimit() uint64 {
 	return context.blockGasLimit
 }
 
-func (context *meteringContext) DeductInitialGasForExecution(input *vmcommon.ContractCallInput, contract []byte) (uint64, error) {
-	remainingGas, err := context.deductInitialGas(
-		input.GasProvided,
+func (context *meteringContext) DeductInitialGasForExecution(contract []byte) error {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/metering.go
```diff
@@ -63,53 +65,88 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
+
+	err = context.deductAndLockGasIfAsyncStep()
+
+	return err
 }
 
-func (context *meteringContext) DeductInitialGasForDirectDeployment(input *vmcommon.ContractCreateInput) (uint64, error) {
-	remainingGas, err := context.deductInitialGas(
-		input.GasProvided,
+func (context *meteringContext) DeductInitialGasForDirectDeployment(input *vmcommon.ContractCreateInput) error {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/contexts/runtime.go
```diff
@@ -41,11 +44,13 @@ func NewRuntimeContext(
 	return context, nil
 }
 
-func (context *runtimeContext) InitState() {
-	context.vmInput = &vmcommon.VMInput{}
-	context.scAddress = make([]byte, 0)
-	context.callFunction = ""
-	context.readOnly = false
+func (runtime *runtimeContext) InitState() {
```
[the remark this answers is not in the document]
camilbancioiu: Indeed. Will not be part of this PR, but the point is good.

arwen/contexts/runtime.go
```diff
@@ -139,17 +145,18 @@ func (context *runtimeContext) Arguments() [][]byte {
...
+	if err != nil {
+		context.host.Output().SetReturnMessage(err.Error())
+	}
+	context.SetRuntimeBreakpointValue(arwen.BreakpointExecutionFailed)
 }
 
 func (context *runtimeContext) SignalUserError(message string) {
-	// SignalUserError() remains in Runtime, and won't be moved into Output,
+	// SignalUserError() remains in runtimeContext, and won't be moved into Output,
 	// because there will be extra handling added here later, which requires
```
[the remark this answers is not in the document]
camilbancioiu: It was actually a TODO, but it was resolved. Comment removed.

arwen/elrondapi/elrondei.go
```diff
@@ -377,13 +383,62 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
...
 	return 0
 }
 
+//export asyncCall
+func asyncCall(context unsafe.Pointer, destOffset int32, valueOffset int32, dataOffset int32, length int32) {
+	runtime := arwen.GetRuntimeContext(context)
+	metering := arwen.GetMeteringContext(context)
+	output := arwen.GetOutputContext(context)
+
+	send := runtime.GetSCAddress()
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/elrondapi/elrondei.go
```diff
@@ -377,13 +383,62 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
...
+	}
+
+	gasSchedule := metering.GasSchedule()
+	gasToUse := gasSchedule.ElrondAPICost.AsyncCallStep
+	gasToUse += gasSchedule.BaseOperationCost.DataCopyPerByte * uint64(length)
+	metering.UseGas(gasToUse)
+
+	gasLimit := metering.GasLeft()
+
+	minAsyncCallCost := 2*gasSchedule.ElrondAPICost.AsyncCallStep + gasSchedule.ElrondAPICost.AsyncCallbackGasLock
```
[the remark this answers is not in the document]
camilbancioiu: I'd rather have it explicit for now.

arwen/elrondapi/elrondei.go
```diff
@@ -377,13 +383,62 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
...
+
+	data, err := runtime.MemLoad(dataOffset, length)
+	if arwen.WithFault(err, context) {
+		return
+	}
+
+	gasSchedule := metering.GasSchedule()
+	gasToUse := gasSchedule.ElrondAPICost.AsyncCallStep
+	gasToUse += gasSchedule.BaseOperationCost.DataCopyPerByte * uint64(length)
+	metering.UseGas(gasToUse)
```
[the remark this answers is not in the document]
camilbancioiu: The same pattern occurs in other API functions as well. We could create appropriate functions for all of them, although having the calculation explicit is also readable. Could be implemented during future refactoring.

arwen/elrondapi/elrondei.go
```diff
@@ -377,13 +383,62 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
...
+
+	gasLimit := metering.GasLeft()
+
+	minAsyncCallCost := 2*gasSchedule.ElrondAPICost.AsyncCallStep + gasSchedule.ElrondAPICost.AsyncCallbackGasLock
+	if gasLimit < minAsyncCallCost {
+		runtime.SetRuntimeBreakpointValue(arwen.BreakpointOutOfGas)
```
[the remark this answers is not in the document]
camilbancioiu: We should. Good catch, this was bad, as the execution would have continued, up to setting the `BreakpointAsyncCall` at the end of the function...

arwen/elrondapi/elrondei.go
```diff
@@ -377,13 +383,62 @@ func transferValue(context unsafe.Pointer, gasLimit int64, destOffset int32, val
...
 }
 
+//export asyncCall
+func asyncCall(context unsafe.Pointer, destOffset int32, valueOffset int32, dataOffset int32, length int32) {
+	runtime := arwen.GetRuntimeContext(context)
+	metering := arwen.GetMeteringContext(context)
+	output := arwen.GetOutputContext(context)
+
+	send := runtime.GetSCAddress()
+	dest, err := runtime.MemLoad(destOffset, arwen.AddressLen)
```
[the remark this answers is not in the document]
camilbancioiu: Yes. Replaced.

arwen/host/breakpoints.go
```diff
@@ -8,19 +8,19 @@ import (
 func (host *vmHost) handleBreakpoint(
 	breakpointValue arwen.BreakpointValue,
 	result wasmer.Value,
-	err error,
 ) error {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/host/execution.go
```diff
@@ -36,14 +33,13 @@ func (host *vmHost) doRunSmartContractCreate(input *vmcommon.ContractCreateInput
 	runtime.SetSCAddress(address)
 	output.AddTxValueToAccount(address, input.CallValue)
 
-	gasForDeployment, err := metering.DeductInitialGasForDirectDeployment(input)
+	err = metering.DeductInitialGasForDirectDeployment(input)
```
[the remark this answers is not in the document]
camilbancioiu: Yes, it's the only one.

arwen/host/execution.go
```diff
@@ -117,43 +113,44 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
 	}()
 
 	result, returnCode, err := host.callSCMethod()
+
+	metering.UnlockGasIfAsyncStep()
```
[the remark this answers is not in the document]
camilbancioiu: Good question. Moved the unlocking to make it more clear - unlocking gas only makes sense on the successful path.

arwen/host/asyncCall.go
camilbancioiu: Will be changed soon. A unit test will be written in the corresponding PR.

arwen/host/asyncCall.go
camilbancioiu: Unit tests are already in development on a separate branch - this function will definitely have its unit tests.

arwen/host/asyncCall.go
camilbancioiu: This could be a change in an upcoming refactoring, together with other instances of this pattern in the API. Noted.

arwen/host/execution.go
```diff
@@ -18,7 +18,10 @@ func (host *vmHost) doRunSmartContractCreate(input *vmcommon.ContractCreateInput
 	var err error
 	defer func() {
 		if err != nil {
-			message := err.Error() + " - " + output.ReturnMessage()
+			message := output.ReturnMessage()
+			if err != arwen.ErrSignalError {
```
andreibancioiu: `If is errSignalError, then ... else ...` may be a bit readable because of the `else`. Just a suggestion.

arwen/host/execution.go
```diff
@@ -77,7 +80,10 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
...
```
andreibancioiu: `If ... else` like above, just a suggestion.

arwen/common.go
```diff
@@ -3,19 +3,27 @@ package arwen
 type StorageStatus int
 
 const (
-	StorageUnchanged StorageStatus = 0
-	StorageModified  StorageStatus = 1
-	StorageAdded     StorageStatus = 3
-	StorageDeleted   StorageStatus = 4
+	StorageUnchanged StorageStatus = iota
```
sasurobert: iota only for the first, the rest can be left unassigned. it does automatically.

arwen/common.go
```diff
@@ -3,19 +3,27 @@ package arwen
...
 type BreakpointValue uint64
 
 const (
-	BreakpointNone        BreakpointValue = 0
-	BreakpointAbort       BreakpointValue = 1
-	BreakpointAsyncCall   BreakpointValue = 2
-	BreakpointSignalError BreakpointValue = 3
-	BreakpointSignalExit  BreakpointValue = 4
-	BreakpointOutOfGas    BreakpointValue = 5
+	BreakpointNone            BreakpointValue = iota
```
sasurobert: same as above.

arwen/contexts/metering.go
```diff
@@ -67,70 +67,65 @@ func (context *meteringContext) BoundGasLimit(value int64) uint64 {
...
-	err = context.deductAndLockGasIfAsyncStep()
-
-	return err
+	return context.deductAndLockGasIfAsyncStep()
 }
 
+// DeductInitialGasForDirectDeployment deducts gas for the deployment of a contract initiated by a Transaction
 func (context *meteringContext) DeductInitialGasForDirectDeployment(input *vmcommon.ContractCreateInput) error {
-	err := context.deductInitialGas(
+	return context.deductInitialGas(
```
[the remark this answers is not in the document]
sasurobert: +1

arwen/helpers.go
```diff
@@ -63,13 +63,15 @@ func InverseBytes(data []byte) []byte {
 	return invBytes
 }
 
-func WithFault(err error, context unsafe.Pointer) bool {
+func WithFault(err error, context unsafe.Pointer, failExecution bool) bool {
```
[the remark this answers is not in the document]
sasurobert: +1

arwen/ethapi/ethereumei.go
```diff
@@ -619,18 +619,18 @@ func ethrevert(context unsafe.Pointer, dataOffset int32, length int32) {
 	metering := arwen.GetMeteringContext(context)
 
 	data, err := runtime.MemLoad(dataOffset, length)
-	if withFault(err, context) {
+	if arwen.WithFault(err, context, false) {
```
sasurobert: ethereum API error handling is still wrong. Right now, as the code is, any execution of the ethereum API consumes all the gasLeft, so there is nothing the developer can do anymore even if the code goes back to the wasmer instance. For those methods where there is no return value, the execution can be stopped. 
For methods which have return 0 / 1 / -1 only the API gas cost has to be deducted, not all the remaining.
In case of revert, there is no need to make any processing. Just use all the gas and runTime.SignalUserError revert.

camilbancioiu: Still wrong, indeed. Fixed as follows:
* `arwen.WithFault()` now only consumes remaining gas if `failExecution == true`
* Ethereum API methods with *no return value* now call `WithFault` with `failExecution == true` 
* Ethereum API methods with *integer return value* now call `WithFault` with `failExecution == false`, leaving the gas unconsumed
* Changed Ethereum API methods to always consume gas first, before any possible failure

arwen/common.go
camilbancioiu: That was the first thing I tried. Got this:
```
arwen/common.go:7:19: const declaration cannot have type without expression
```

camilbancioiu: Nevermind, the problem was the remaining datatype. Fixed.

arwen/common.go
camilbancioiu: Fixed.

arwen/host/execution.go
camilbancioiu: Changed.

arwen/host/execution.go
camilbancioiu: Changed.

arwen/contexts/metering.go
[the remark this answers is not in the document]
camilbancioiu: `deductInitialGas` is called in three methods of `meteringContext`: once with `CompilePerByte`, twice with `StorePerByte`, depending on the caller. Should keep the third argument for now.

arwen/helpers.go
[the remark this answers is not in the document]
camilbancioiu: Changed accordingly.

arwen/helpers.go
```diff
@@ -61,3 +62,20 @@ func InverseBytes(data []byte) []byte {
...
+	if failExecution {
+		runtime := GetRuntimeContext(context)
+		metering := GetMeteringContext(context)
+
+		metering.UseGas(metering.GasLeft())
+		runtime.FailExecution(err)
+	}
+
+	return true
+
```
andreibancioiu: Remove empty line.

arwen/ethapi/ethereumei.go
```diff
@@ -618,19 +619,19 @@ func ethrevert(context unsafe.Pointer, dataOffset int32, length int32) {
...
 
 	output.ClearReturnData()
 	output.Finish(data)
-	// TODO replace this with a breakpoint?
-	runtime.SignalUserError("revert")
 
-	gasToUse := metering.GasSchedule().EthAPICost.Revert
-	gasToUse += metering.GasSchedule().BaseOperationCost.PersistPerByte * uint64(length)
-	metering.UseGas(gasToUse)
+	runtime.SignalUserError("revert")
```
andreibancioiu: `revert` seems correct. @sasurobert, ok?

arwen/ethapi/ethereumei.go
andreibancioiu: Revert seems correct to me now as well.

arwen/ethapi/ethereumei.go
```diff
@@ -710,12 +711,12 @@ func ethreturnDataCopy(context unsafe.Pointer, resultOffset int32, dataOffset in
 	}
 
 	returnDataSlice, err := arwen.GuardedGetBytesSlice(ethReturnData, dataOffset, length)
```
andreibancioiu: Just a question:

```
if int32(len(ethReturnData)) < dataOffset+length {
	return
}
```

Is this desired behavior or a silent error? If it weren't this return here, then the subsequent `arwen.GuardedGetBytesSlice` would have resulted in error (bad bounds), then fault.

Don't we actually want to fault here instead of hiding the error and return to the caller and don't even tell him that the copy didn't work? @sasurobert

arwen/ethapi/ethereumei.go
andreibancioiu: Extra question here, @camilbancioiu, @sasurobert.
According to the [docs](https://ewasm.readthedocs.io/en/mkdocs/eth_interface/), some functions return 2 on revert:

```
result: i32 - Returns 0 on success, 1 on failure and 2 on revert
```

But we do not have any `return 2` in our `ethereumei.go` - are we missing something?

sasurobert: we have to return as those docs say

arwen/ethapi/ethereumei.go
sasurobert: good catch. We should have an error here.

arwen/ethapi/ethereumei.go
sasurobert: signal error can be called directly from the first line. All the data is thrown out in any case.

arwen/ethapi/ethereumei.go
camilbancioiu: Seems like need a full conformity check on our Ethereum API, but that's beyond this PR. Created a JIRA issue for it: https://elrondnetwork.atlassian.net/browse/EN-5772

arwen/ethapi/ethereumei.go
camilbancioiu: Indeed. Added a call to `arwen.WithFault()` before return.

arwen/ethapi/ethereumei.go
camilbancioiu: Moving `SignalUserError()` above `WithFault()` will not change anything (they will override each other completely), but the function is more readable with `SignalUserError()` at the end.

arwen/helpers.go
camilbancioiu: Done.

PR #71, opened 2020-02-13
Add checks upon contract deployment

andreibancioiu:
- Check that memory declaration exists in the wasm bytecode.
 - Check that the exported functions have ASCII names.
 - Check that the exported functions do not have reserved names.

### DIFF
--- a/arwen/contexts/guard.go
+++ b/arwen/contexts/guard.go
@@ -5,14 +5,33 @@ import (
 	"unicode"
 
 	"github.com/ElrondNetwork/arwen-wasm-vm/arwen"
+	"github.com/ElrondNetwork/arwen-wasm-vm/wasmer"
 )
 
 func (context *runtimeContext) VerifyContractCode() error {
-	if !context.instance.HasMemory() {
+	err := verifyMemoryDeclaration(context.instance)
+	if err != nil {
+		return err
+	}
+
+	err = verifyFunctionsNames(context.instance)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func verifyMemoryDeclaration(instance *wasmer.Instance) error {
+	if !instance.HasMemory() {
 		return arwen.ErrMemoryDeclarationMissing
 	}
 
-	for functionName := range context.instance.Exports {
+	return nil
+}
+
+func verifyFunctionsNames(instance *wasmer.Instance) error {
+	for functionName := range instance.Exports {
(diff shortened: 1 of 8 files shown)


arwen/host/execution.go
```diff
@@ -209,7 +215,12 @@ func (host *vmHost) CreateNewContract(input *vmcommon.ContractCreateInput) ([]by
 
 	err = runtime.CreateWasmerInstance(input.ContractCode, gasForDeployment)
 	if err != nil {
-		fmt.Println("arwen Error: ", err.Error())
+		output.Transfer(input.CallerAddr, address, 0, input.CallValue, nil)
```
sasurobert: we can take this code out output.Transfer in case there is an error. If the any vmexecution returns with error, only the error code and the returnData is taken into consideration by the smart contract processor.

arwen/contexts/guard.go
```diff
@@ -0,0 +1,69 @@
...
+	return nil
+}
+
+func isValidFunctionName(functionName string) bool {
+	const maxLengthOfFunctionName = 256
+
+	if len(functionName) == 0 {
+		return false
+	}
+
```
sasurobert: delete empty line

arwen/contexts/guard.go
```diff
@@ -0,0 +1,69 @@
...
+	if len(functionName) >= maxLengthOfFunctionName {
+		return false
+	}
+
```
sasurobert: delete empty line

sasurobert: we might add a few functions will be unable to be called, builtin elrond functions. I would add in arwen constructor a map[string]struct{} which function names are taken. One example right now is "claimDeveloperRewards"

andreibancioiu: I've added a new struct called `ReservedFunctions`. If this is ok, I can move it to `vm-common`, so that the protocol can also reference it in a clean manner.

arwen/contexts/guard.go
andreibancioiu: Done.

arwen/host/execution.go
andreibancioiu: Fixed, we do not undo the transfer anymore - since the transfer will not take place anyway. Correct?

sasurobert: correct. it is not saved in the vmoutput.

PR #77, opened 2020-02-24
Byte inversion fix

camilbancioiu:


arwen/elrondapi/elrondei.go
```diff
@@ -891,8 +888,7 @@ func executeOnSameContext(
 	gasToUse += metering.GasSchedule().BaseOperationCost.DataCopyPerByte * uint64(actualLen)
 	metering.UseGas(gasToUse)
 
-	invBytes := arwen.InverseBytes(value)
-	bigIntVal := big.NewInt(0).SetBytes(invBytes)
+	bigIntVal := big.NewInt(0).SetBytes(value)
```
sasurobert: I am not sure that these api's would work without inversion.

camilbancioiu: Added tests specifically for `getCallValue()` and `int64getArguments()`. Tests for `executeOn(Same|Dest)Context()` are on the execution tests branch.

arwen/host/elrondei_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	// 1-byte call value
+	host, _ := DefaultTestArwenForCall(t, code)
+	input := DefaultTestContractCallInput()
+	input.GasProvided = 100000
+	input.Function = "test_getCallValue_1byte"
+	input.CallValue = big.NewInt(64)
+
+	vmOutput, err := host.RunSmartContractCall(input)
+	assert.Nil(t, err)
```
andreibancioiu: Use `require` instead of `assert`.

test/contracts/elrondei/elrondei.c
```diff
@@ -0,0 +1,88 @@
...
+	} else {
+		finish(msg_not_ok, 6);
+	}
+
+	finish((byte*)&length, 4);
+	finish(value, length);
+}
+
+void test_getCallValue_bigInt_to_Bytes() {
+	int length = getCallValue(value);
```
andreibancioiu: Add extra test to compare (assert equality) the value from `bigIntGetCallValue` and a big int obtained initialized with the bytes from `getCallValue`.

test/contracts/elrondei/elrondei.c
```diff
@@ -0,0 +1,88 @@
+#include "../elrond/context.h"
+
```
andreibancioiu: Add extra test to call `executeOnSameContext` to another function in another contract, function which takes the call value and calls `transferValue` / 2 to some hard-coded address.

camilbancioiu: Will come as a next PR, after `executeOnSameContext` is fixed.

arwen/helpers.go
```diff
@@ -50,6 +50,23 @@ func GuardedGetBytesSlice(data []byte, offset int32, length int32) ([]byte, erro
...
+	if len(data) == 0 {
+		return []byte{}
+	}
+	padSize := size - len(data)
+	if padSize <= 0 {
+		return data
+	}
+
+	paddedBytes := make([]byte, padSize)
+	paddedBytes = append(paddedBytes, data...)
```
sasurobert: why is this used ?

PR #79, opened 2020-02-27
Async test

andrei-marinica:
Contains:
  - async call tests (with callbacks)
  - removed possibility to return anything (other than void) directly from wasmer


arwen/host/execution.go
```diff
@@ -303,13 +299,14 @@ func (host *vmHost) execute(input *vmcommon.ContractCallInput) error {
 		return arwen.ErrFunctionRunError
 	}
 
+	if !result.IsVoid() {
```
camilbancioiu: It might be too restrictive to throw an error if the called function returns a value. Better to simply ignore `result` altogether?

andrei-marinica: I think it is better to signal to the developer that the return value is not allowed than ignoring it and making them wonder why what they thought should work doesn't work.

It should only be an issue for developers going 100% raw - even for C I'm thinking we should come up with some macro magic.

camilbancioiu:
All tests have passed on my machine.

PR #80, opened 2020-02-27
Added libwasmer_runtime_c_api.so arm64 version

groninge01:


groninge01:
I have added the "arm64" tag to the filename because I don't want to overwrite the existing one.

PR #87, opened 2020-03-10
Fix WASM memory deallocation

camilbancioiu:
Fixes a memory leak that was caused by not deallocating the WASM memory of a Wasmer instance.


arwen/contexts/runtime.go
```diff
@@ -113,6 +113,11 @@ func (context *runtimeContext) PopInstance() {
 }
 
 func (context *runtimeContext) ClearInstanceStack() {
+	if len(context.instanceStack) > 0 {
```
sasurobert: no need for this if. going directly over the range is good enough.

camilbancioiu: Fixed.

PR #92, opened 2020-03-11
Fix after review

camilbancioiu:


arwen/contexts/output.go
```diff
@@ -150,14 +150,14 @@ func (context *outputContext) WriteLog(address []byte, topics [][]byte, data []b
 // Transfer handles any necessary value transfer required and takes
 // the necessary steps to create accounts and reverses the state in case of an
 // execution error or failed value transfer.
-func (context *outputContext) Transfer(destination []byte, sender []byte, gasLimit uint64, value *big.Int, input []byte) int {
-	if value.Cmp(big.NewInt(0)) < 0 {
-		return 1
+func (context *outputContext) Transfer(destination []byte, sender []byte, gasLimit uint64, value *big.Int, input []byte) error {
+	zero := big.NewInt(0)
```
sasurobert: this zero can be move as a global variable.

PR #93, opened 2020-03-16
Wasmer benchmarks

camilbancioiu:
This PR adds the ERC20 benchmark from `elrond-go` into Arwen, so it can be run independently of the Node. The `BlockchainHookMock` has been adapted to apply the `VMOutput` to its internal storage of mock `Account`s, emulating the Node's state.


arwen/host/execution_benchmark_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	"github.com/ElrondNetwork/arwen-wasm-vm/arwen"
+	"github.com/ElrondNetwork/arwen-wasm-vm/config"
+	"github.com/ElrondNetwork/arwen-wasm-vm/mock"
+	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
+	"github.com/stretchr/testify/require"
+)
+
+var owner = []byte("owner")
+var receiver = []byte("receiver")
```
andreibancioiu: This means that the test involves only 2 accounts. But since the benchmark's purpose is mostly to check wasmer performance (upgrade / downgrade), it should be sufficient.

arwen/host/execution_benchmark_test.go
```diff
@@ -0,0 +1,123 @@
...
+	require.Equal(tb, vmcommon.Ok, vmOutput.ReturnCode)
+
+	// Ensure the deployment persists in the mock BlockchainHook
+	mockBlockchainHook.UpdateAccounts(vmOutput.OutputAccounts)
+	return host, mockBlockchainHook
+}
+
+func verifyTransfers(tb testing.TB, mockBlockchainHook *mock.BlockchainHookMock, totalTokenSupply *big.Int) {
+	ownerKey := string([]byte{
+		1, 0, 'o', 'w', 'n', 'e', 'r', 0,
```
andreibancioiu: Not sure why the `1, 0` prefix. Can you also create a new helper function for such tests? `createStorageKey()` that does the padding?

mock/blockChainHookMock.go
```diff
@@ -237,3 +242,42 @@ func (b *BlockchainHookMock) CurrentRandomSeed() []byte {
...
+		account, exists := b.Accounts[strAddress]
+		if !exists {
+			account = &Account{
+				Exists:  true,
+				Address: outputAccount.Address,
+				Balance: big.NewInt(0).Set(outputAccount.BalanceDelta),
+				Code:    outputAccount.Code,
+				Storage: make(map[string][]byte),
+			}
+		} else {
```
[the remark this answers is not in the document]
camilbancioiu: Done.

arwen/host/execution_benchmark_test.go
```diff
@@ -0,0 +1,123 @@
...
+	verifyTransfers(tb, mockBlockchainHook, totalTokenSupply)
+}
+
+func deploy(tb testing.TB, totalTokenSupply *big.Int) (*vmHost, *mock.BlockchainHookMock) {
+	// Prepare the host
+	mockBlockchainHook := mock.NewBlockchainHookMock()
+	mockBlockchainHook.AddAccount(&mock.Account{
+		Address: owner,
+		Nonce:   1024,
+		Balance: big.NewInt(88000),
```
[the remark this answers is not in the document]
camilbancioiu: Changed to `0`. It was a meaningless number, indeed.

arwen/host/execution_benchmark_test.go
camilbancioiu: Yes, just 2 accounts should be enough for checking performance. We could add more accounts in the future, for other scenarios.

arwen/host/execution_benchmark_test.go
camilbancioiu: Created a function which generates these keys. It's how the ERC20 expects them, but don't know the reason for the prefix right now.

PR #96, opened 2020-03-18
Circumvent Go's GC

andreibancioiu:


Makefile
```diff
@@ -1,10 +1,16 @@
...
 	go build ./...
 
-test:
+build-arwen:
+	go build -o ./cmd/arwen/arwen ./cmd/arwen
+	cp ./cmd/arwen/arwen ./ipc/tests
+
+test: build-arwen run-tests
+
+run-tests:
```
camilbancioiu: Add dependency on `clean` as well (i.e. `run-tests: clean`). I forgot to add it to `test`.

Makefile
```diff
@@ -1,10 +1,16 @@
-.PHONY: test build clean
+.PHONY: test build build-arwen clean
```
camilbancioiu: `run-tests` should be added here as well. All `make` targets are phony, unless they are real filenames.

andreibancioiu: Added.

Makefile
andreibancioiu: Added.

PR #104, opened 2020-03-22
Extra logs and measurements for Arwen OOP

andreibancioiu:
Separate loggers: driver logger (Node's process), main Arwen logger (Arwen's process), dialogue logger (both processes).

Added counts for deployed and executed (not kept between restarts, of course).

Added prints for Wasmer compilation phase (will be redirected to Arwen's main logger, with Trace).

### DIFF
--- a/arwen/contexts/runtime.go
+++ b/arwen/contexts/runtime.go
@@ -2,6 +2,7 @@ package contexts
 
 import (
 	"fmt"
+	"time"
 	"unsafe"
 
 	"github.com/ElrondNetwork/arwen-wasm-vm/arwen"
@@ -60,6 +61,8 @@ func (context *runtimeContext) InitState() {
 }
 
 func (context *runtimeContext) CreateWasmerInstance(contract []byte, gasLimit uint64) error {
+	defer arwen.TimeTrack(time.Now(), "CreateWasmerInstance")
+
 	var err error
 	context.instance, err = wasmer.NewMeteredInstance(contract, gasLimit)
 	if err != nil {
--- a/arwen/contexts/runtime.go
+++ b/arwen/contexts/runtime.go
@@ -2,7 +2,6 @@ package contexts
 
 import (
 	"fmt"
-	"time"
 	"unsafe"
 
 	"github.com/ElrondNetwork/arwen-wasm-vm/arwen"
@@ -61,8 +60,6 @@ func (context *runtimeContext) InitState() {
 }
 
 func (context *runtimeContext) CreateWasmerInstance(contract []byte, gasLimit uint64) error {
-	defer arwen.TimeTrack(time.Now(), "CreateWasmerInstance")
-
 	var err error
 	context.instance, err = wasmer.NewMeteredInstance(contract, gasLimit)
 	if err != nil {
--- a/arwen/helpers.go
+++ b/arwen/helpers.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"math/big"
 	"path/filepath"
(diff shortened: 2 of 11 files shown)


arwen/contexts/runtime.go
```diff
@@ -60,6 +61,8 @@ func (context *runtimeContext) InitState() {
 }
 
 func (context *runtimeContext) CreateWasmerInstance(contract []byte, gasLimit uint64) error {
+	defer arwen.TimeTrack(time.Now(), "CreateWasmerInstance")
```
camilbancioiu: This will get called a lot... `TimeTrack` prints directly to `stdout`, consuming time and filling up the terminal. Do we want it here? It is useful for benchmarking, but not as a permanent feature.

arwen/helpers.go
```diff
@@ -86,3 +87,8 @@ func GetSCCode(fileName string) []byte {
 
 	return code
 }
+
+func TimeTrack(start time.Time, message string) {
+	elapsed := time.Since(start)
+	fmt.Println(message, "duration", elapsed)
```
camilbancioiu: Do we have access to a logger here, for `logger.Trace` instead of `fmt.Println`?

arwen/contexts/runtime.go
andreibancioiu: Not a permanent feature. This will write to Arwen's stdout, which is captured and displayed with log.Trace() at Node's side. Indeed, it will fill up the terminal during benchmarks. Removed it for now, we will add it again perhaps when Arwen will reference the "true" logger component.

arwen/helpers.go
andreibancioiu: No, we don't at this moment. Only when Arwen will reference the "true" logger.

ipc/nodepart/arwenDriver.go
```diff
@@ -230,7 +260,8 @@ func (driver *ArwenDriver) IsClosed() bool {
 
 // RunSmartContractCreate sends a deploy request to Arwen and waits for the output
 func (driver *ArwenDriver) RunSmartContractCreate(input *vmcommon.ContractCreateInput) (*vmcommon.VMOutput, error) {
-	driver.nodeLogger.Trace("RunSmartContractCreate")
+	driver.counterDeploy++
```
sasurobert: this might overflow.

ipc/nodepart/arwenDriver.go
```diff
@@ -348,15 +381,25 @@ func (driver *ArwenDriver) continuouslyCopyArwenLogs(arwenStdout io.Reader, arwe
...
 
 			line = strings.TrimSpace(line)
-			driver.nodeLogger.Error("ARWEN-ERR", "line", line)
+			driver.arwenMainLogger.Error("ARWEN-ERR", "line", line)
+		}
+	}()
+
+	go func() {
+		for {
+			err := logger.ReceiveLogThroughPipe(driver.arwenMainLogger, arwenLog, driver.logsMarshalizer)
```
sasurobert: duplicated code - move to a separate func

ipc/nodepart/arwenDriver.go
andreibancioiu: Only after Arwen processes 18,446,744,073,709,551,615 smart contracts without any restart :) With 1000 SC per second, this will be in ~ 584942417 years :)

ipc/nodepart/arwenDriver.go
andreibancioiu: Fixed.

PR #107, opened 2020-04-03
Pass ProtocolReservedFunctions to Arwen

andreibancioiu:
Pass `ProtocolReservedFunctions` (eg. `claimDeveloperRewards`) to Arwen (from Node).


arwen/host/arwen.go
```diff
@@ -42,7 +42,8 @@ func NewArwenVM(
 	cryptoHook vmcommon.CryptoHook,
 	vmType []byte,
 	blockGasLimit uint64,
-	gasSchedule map[string]map[string]uint64,
+	gasSchedule config.GasScheduleMap,
+	protocolReservedFunctions []string,
```
sasurobert, line 46: can you make the input as the gasScheduleMap. a defined type, as we will have reservedFunctions, builtInFunctions and maybe others ?

arwen/contexts/runtime.go
```diff
@@ -33,10 +33,7 @@ type runtimeContext struct {
 }
 
 // NewRuntimeContext creates a new runtimeContext
-func NewRuntimeContext(
-	host arwen.VMHost,
-	vmType []byte,
-) (*runtimeContext, error) {
+func NewRuntimeContext(host arwen.VMHost, vmType []byte, protocolReservedFunctions []string) (*runtimeContext, error) {
```
camilbancioiu, line 36: I suggest not passing `protocolReservedFunctions` as argument directly to this constructor, but instead the constructor should retrieve these function names from the `host`, with a getter. See `host.GetAPIMethods()` and its use in `NewRuntimeContext()` as an example.

ipc/common/arguments.go
```diff
@@ -18,14 +19,12 @@ type ArwenArguments struct {
 
 // VMHostArguments represents the arguments to be passed to VMHost
 type VMHostArguments struct {
-	VMType        []byte
-	BlockGasLimit uint64
-	GasSchedule   GasScheduleMap
+	VMType                    []byte
```
camilbancioiu, line 22: The `VMHostArguments` structure could be moved in `arwen/common.go`, if possible, and then `NewArwenVM()` can be changed to take it as argument, instead of the individual values.

arwen/host/arwen.go
andreibancioiu: Fixed, `vmcommon.FunctionNames`.

arwen/contexts/runtime.go
andreibancioiu: Fixed.

ipc/common/arguments.go
andreibancioiu: Fixed (and renamed).

arwen/host/arwen.go
```diff
@@ -35,27 +35,27 @@ type vmHost struct {
...
-		blockChainHook:    blockChainHook,
-		cryptoHook:        cryptoHook,
-		meteringContext:   nil,
-		runtimeContext:    nil,
-		blockchainContext: nil,
-		storageContext:    nil,
-		bigIntContext:     nil,
-		scAPIMethods:      nil,
+		blockChainHook:           blockChainHook,
+		cryptoHook:               cryptoHook,
```
camilbancioiu, line 51: Please add checks against `nil` values for the received `blockChainHook` and `cryptoHook`, so the constructor can return an error. I forgot to add them previously.

PR #109, opened 2020-04-07
Use elrond logger in Arwen

andreibancioiu:
- Removed pipe logging components from Arwen
 - Referenced the logging components from `elrond-go-logger`
 - Updated readme


ipc/arwenpart/messenger.go
```diff
@@ -4,19 +4,21 @@ import (
 	"os"
 
 	"github.com/ElrondNetwork/arwen-wasm-vm/ipc/common"
-	"github.com/ElrondNetwork/arwen-wasm-vm/ipc/logger"
 	"github.com/ElrondNetwork/arwen-wasm-vm/ipc/marshaling"
+	logger "github.com/ElrondNetwork/elrond-go-logger"
 )
 
+var messengerLog = logger.GetOrCreate("arwen/messenger")
```
camilbancioiu, line 11: Can `messengerLog` be renamed to `log`, like in the other files?

andreibancioiu: I wanted to have separate loggers for `arwen/part` and `arwen/messenger`.

andreibancioiu: Fixed :)

PR #114, opened 2020-04-13
Execute builtin functions via the blockchain hook, when called by SCs indirectly

camilbancioiu:


arwen/host/execution_test.go
```diff
@@ -740,6 +740,80 @@ func TestExecution_ExecuteOnSameContext_Recursive_Mutual_SCs_OutOfGas(t *testing
...
+
+		var empty struct{}
+		names["builtinClaim"] = empty
+		names["builtinDoSomething"] = empty
+		return names
+	}
+
+	host, stubBlockchainHook := DefaultTestArwenForCall(t, code)
+	stubBlockchainHook.GetBalanceCalled = getBalanceCalled
+	// stubBlockchainHook.GetBuiltinFunctionNamesCalled = getBuiltinFunctionNames
```
andreibancioiu, line 765: Remove this?

arwen/host/execution.go
```diff
@@ -411,6 +421,21 @@ func (host *vmHost) callSCMethodIndirect() error {
...
+	_, _, metering, output, runtime, _ := host.GetContexts()
+
+	valueToSend, gasConsumed, err := host.blockChainHook.ProcessBuiltInFunction(input)
+	if err != nil {
+		metering.UseGas(gasConsumed)
+		return err
+	}
+
+	scAccount, _ := output.GetOutputAccount(runtime.GetSCAddress())
+	scAccount.BalanceDelta.Add(scAccount.BalanceDelta, valueToSend)
```
andreibancioiu, line 434: Just a question - this is always addition, correct? Or perhaps `valueToSend` could also be negative?

sasurobert: it does not matter. add can do minus as well. built-in functions is has a tons of checks.

arwen/host/execution.go
```diff
@@ -411,6 +421,21 @@ func (host *vmHost) callSCMethodIndirect() error {
...
+func (host *vmHost) callBuiltinFunction(input *vmcommon.ContractCallInput) error {
+	_, _, metering, output, runtime, _ := host.GetContexts()
+
+	valueToSend, gasConsumed, err := host.blockChainHook.ProcessBuiltInFunction(input)
+	if err != nil {
+		metering.UseGas(gasConsumed)
+		return err
+	}
+
+	scAccount, _ := output.GetOutputAccount(runtime.GetSCAddress())
```
sasurobert, line 433: seeing this I think we have to make sure who gets the money, now it is correct, as only in claimdeveloperrewards - and the sender can get it. however processbuiltinfunction might have to return a vmoutput ? not know, but we should discuss it

arwen/host/execution_test.go
sasurobert: remove it :D

arwen/host/execution.go
camilbancioiu: If a builtin function transfers money that aren't necessarily destined to the caller of the function, then yes, more info must be returned to the VM. The cleanest alternative I can think of right now is to return a map of `OutputAccount` objects, which can be merged by the VM on the `outputContext`. This might be better than a full `VMOutput`, although `ReturnData` would also be very useful. Yes, this needs more discussion.

arwen/host/execution_test.go
camilbancioiu: Removed.

PR #115, opened 2020-04-13
Delegation and DNS tests

andrei-marinica:
Contains:
 - delegation tests
 - DNS tests
 - other updated tests
 - some minor fixes in int64getArgument and int64finish


arwen/errorUserMsg.go
```diff
@@ -0,0 +1,5 @@
+package arwen
+
+const UserErrorArgIndexOutOfRange = "argument index out of range"
```
andreibancioiu, line 3: `bigIntOps` etc. contain some hardcoded error messages. Perhaps you can also move the here? Can be done in another PR, of course.

PR #121, opened 2020-05-11
Api features test 4

andrei-marinica:
Contains:
- sending errors as arguments to callback
- block info in tests
- delegation unstake, unbond, force unstake


arwen/host/execution.go
```diff
@@ -424,10 +424,10 @@ func (host *vmHost) callSCMethodIndirect() error {
 
 	_, err = function()
 	if err != nil {
-		return arwen.ErrFunctionRunError
+		err = host.handleBreakpointIfAny(err)
```
camilbancioiu, line 427: I've always postponed doing this change, due to testing complexity. It is indeed required to handle breakpoints here, but there's probably 200 edge cases to watch out for. Adding a JIRA task to write tests for it...

config/gasSchedule.go
```diff
@@ -174,7 +174,7 @@ func FillGasMap_ElrondAPICosts(value uint64) map[string]uint64 {
 	gasMap["DelegateExecution"] = value
 	gasMap["ExecuteReadOnly"] = value
 	gasMap["AsyncCallStep"] = value
-	gasMap["AsyncCallbackGasLock"] = value
+	gasMap["AsyncCallbackGasLock"] = 100_000
```
camilbancioiu, line 177: I thought you could only use `_` to separate zeros in Rust. Good to know.

config/gasSchedule.go
```diff
@@ -174,7 +176,7 @@ func FillGasMap_ElrondAPICosts(value uint64) map[string]uint64 {
 	gasMap["DelegateExecution"] = value
 	gasMap["ExecuteReadOnly"] = value
 	gasMap["AsyncCallStep"] = value
-	gasMap["AsyncCallbackGasLock"] = value
+	gasMap["AsyncCallbackGasLock"] = AsyncCallbackGasLock
```
camilbancioiu, line 179: Replace the global constant with a parameter to `FillGasMap_ElrondAPICosts()`, and to `FillGasMap()` and `MakeGasMap()` as well. `MakeGasMap()` is also used in some integration tests in `elrond-go`, where it will need to be updated, but it's better anyway, because there was no way to customize it previously.

An alternative would be to remove the global, add _no_ new parameter, but instead to change the value of `AsyncCallbackGasLock` in the map returned by `MakeGasMap()` in the tests that need it. If it's sufficiently easy, I recommend this alternative, as opposed to the global constant and to the new parameter. Then there's no update needed in `elrond-go`.

arwen/host/execution_test_vmoutputs.go
```diff
@@ -888,7 +888,7 @@ func expectedVMOutput_AsyncCall_CallBackFails() *vmcommon.VMOutput {
 	AddFinishData(vmOutput, []byte{3})
 	AddFinishData(vmOutput, []byte("thirdparty"))
 	AddFinishData(vmOutput, []byte("vault"))
-	AddFinishData(vmOutput, []byte("execution failed"))
+	AddFinishData(vmOutput, []byte("user error"))
```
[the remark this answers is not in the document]
andrei-marinica: Yes, you get the specific error from the async called function in the callback, instead of the general "execution failed". The error message can be used in the callback.

arwen/host/asyncCall.go
```diff
@@ -128,17 +128,22 @@ func (host *vmHost) createCallbackContractCallInput(destinationVMOutput *vmcommo
...
+		big.NewInt(int64(destinationVMOutput.ReturnCode)).Bytes(),
+	}
+	if destinationErr == nil {
+		// when execution went Ok,
+		// use called function results as callback arguments
+		arguments = append(arguments, destinationVMOutput.ReturnData...)
+	} else {
+		// when execution returned error,
+		// also provide error message as argument
+		arguments = append(arguments, []byte(destinationVMOutput.ReturnMessage))
```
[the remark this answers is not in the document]
andrei-marinica: I don't think there's ever too many comments, especially since we don't yet have a complete API documentation ...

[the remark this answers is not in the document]
andrei-marinica: I made them a bit more concise

config/gasSchedule.go
```diff
@@ -725,3 +723,10 @@ func FillGasMap_WASMOpcodeValues(value uint64) map[string]uint64 {
 
 	return gasMap
 }
+
+const GasValueForTests = 1
+const AsyncCallbackGasLockForTests = 100_000
+
+func MakeGasMapForTests() GasScheduleMap {
+	return MakeGasMap(GasValueForTests, AsyncCallbackGasLockForTests)
```
camilbancioiu, line 731: Either put the two global constants at the top of the file, or add them as local constants within the `MakeGasMapForTests()` function.

PR #123, opened 2020-05-14
Api features test 5

andrei-marinica:
Contains:
- user error when attempting to write under elrond reserved key
- new delegation contract (using ELRONDreward) + tests
- json tests support ELRONDreward


arwen/elrondapi/bigintOps.go
```diff
@@ -289,6 +289,11 @@ func bigIntStorageStoreUnsigned(context unsafe.Pointer, keyOffset int32, keyLeng
 		return 0
 	}
 
+	if isElrondReservedKey(key) {
```
camilbancioiu, line 292: This verification is best done in the `storageContext.SetStorage`, for all storage writes, because it's a universal restriction.

arwen/elrondapi/elrondei.go
```diff
@@ -506,6 +507,10 @@ func getNumArguments(context unsafe.Pointer) int32 {
 	return int32(len(args))
 }
 
+func isElrondReservedKey(key []byte) bool {
```
camilbancioiu, line 510: Move as a public method of `storageContext`.

arwen/elrondapi/bigintOps.go
sasurobert: +1

arwen/elrondapi/elrondei.go
sasurobert: +1

arwen/elrondProtectedStorage.go
```diff
@@ -0,0 +1,7 @@
+package arwen
+
+// ElrondProtectedKeyPrefix prefixes all Elrond reserved storage. Only the protocol can write to keys starting with this.
+const ElrondProtectedKeyPrefix = "ELROND"
```
sasurobert, line 4: this could come as input for arwen.

sasurobert: as of protected functions - protected keys.

arwen/elrondProtectedStorage.go
```diff
@@ -0,0 +1,7 @@
...
+
+// ElrondRewardKey is the storage key where the protocol writes when sending out rewards.
+const ElrondRewardKey = ElrondProtectedKeyPrefix + "reward"
```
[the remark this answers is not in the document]
sasurobert: the dApp developer has to known about this key.

arwen/elrondapi/bigintOps.go
sasurobert: +1

arwen/elrondapi/elrondei.go
sasurobert: +1

arwen/elrondProtectedStorage.go
camilbancioiu: Indeed. Created JIRA task for this.

arwen/contexts/storage.go
```diff
@@ -78,9 +85,17 @@ func (context *storageContext) GetStorage(key []byte) []byte {
...
 }
 
-func (context *storageContext) SetStorage(key []byte, value []byte) int32 {
+func (context *storageContext) isElrondReservedKey(key []byte) bool {
+	return bytes.HasPrefix(key, []byte(context.elrondProtectedKeyPrefix))
+}
+
+func (context *storageContext) SetStorage(key []byte, value []byte) (arwen.StorageStatus, error) {
+	if context.isElrondReservedKey(key) {
+		return arwen.StorageUnchanged, errors.New(arwen.UserErrorStoreElrondReservedKey)
```
camilbancioiu, line 94: Create specific error constant in `arwen/errors.go` for this case, and return it here instead of `errors.New(...)`. If you need the message itself in some other place, it's better to take it from within the error, rather than define the message separately then instantiating new errors from it every time.

arwen/elrondapi/elrondei.go
```diff
@@ -535,7 +525,13 @@ func storageStore(context unsafe.Pointer, keyOffset int32, keyLength int32, data
 	gasToUse := metering.GasSchedule().ElrondAPICost.StorageStore
 	metering.UseGas(gasToUse)
 
-	return storage.SetStorage(key, data)
+	storageStatus, err := storage.SetStorage(key, data)
+	if err != nil {
+		runtime.SignalUserError(err.Error())
```
camilbancioiu, line 530: Use `arwen.WithFault()` here, instead of calling `runtime.SignalUserError()` directly. And `storage.SetStorage()` tends to fail due to a SC programming error (forbidden key accessed), so it's not a user error, but an execution failure instead (which makes `arwen.WithFault()` more appropriate).

In this specific case, the last argument to `arwen.WithFault()`, namely `failExecution`, should still be `runtime.ElrondAPIErrorShouldFailExecution()`, currently configured to `true`, because an error in `storage.SetStorage()` should be unrecoverable and cannot be handled by the SmartContract (unless, later, we allow an "expert" mode where errors are caught and handled by the SC, as discussed some months ago).

arwen/elrondapi/elrondei.go
```diff
@@ -826,17 +822,18 @@ func int64storageStore(context unsafe.Pointer, keyOffset int32, keyLength int32,
...
-
 	data := big.NewInt(value)
 
 	gasToUse := metering.GasSchedule().ElrondAPICost.Int64StorageStore
 	metering.UseGas(gasToUse)
 
-	return storage.SetStorage(key, data.Bytes())
+	storageStatus, err := storage.SetStorage(key, data.Bytes())
+	if err != nil {
+		runtime.SignalUserError(err.Error())
```
camilbancioiu, line 832: As above, it's better to use `arwen.WithFault()`.

arwen/contexts/storage_test.go
```diff
@@ -197,10 +197,10 @@ func TestStorageContext_SetStorage(t *testing.T) {
 	key = []byte("RESERVEDkey")
 	value = []byte("doesn't matter")
 	storageStatus, err = storageContext.SetStorage(key, value)
-	require.Equal(t, errors.New(arwen.UserErrorStoreElrondReservedKey), err)
+	require.Equal(t, errors.New(arwen.ErrStoreElrondReservedKey.Error()), err)
```
camilbancioiu, line 200:
```suggestion
	require.Equal(t, arwen.ErrStoreElrondReservedKey, err)
```

arwen/contexts/storage_test.go
```diff
@@ -197,10 +197,10 @@ func TestStorageContext_SetStorage(t *testing.T) {
...
 
 	key = []byte("RESERVED")
 	value = []byte("doesn't matter")
 	storageStatus, err = storageContext.SetStorage(key, value)
-	require.Equal(t, errors.New(arwen.UserErrorStoreElrondReservedKey), err)
+	require.Equal(t, errors.New(arwen.ErrStoreElrondReservedKey.Error()), err)
```
camilbancioiu, line 205: Same as above.

PR #124, opened 2020-06-01
Wasmer options

camilbancioiu:
Update `libwasmer` binaries after disabling memory exception annotations in the Singlepass compiler.

Create `struct CompilationOptions` which now contains the `GasLimit` for the Wasmer instance to be created, along with 3 more boolean options:
* `OpcodeTrace`, default `false`, which causes Wasmer to generate a file called `opcode.trace` in the current directory and to print information on `stdout` about the imports and exports of the compiled WASM module after execution. Most importantly, Wasmer will print to `stdout` the position of the last known executed WASM opcode, position which can be searched for in `opcode.trace` to find the actual last opcode. Enabling `OpcodeTrace` slows down execution. It is intended for debugging only.
* `Metering`, default `true`, which causes Wasmer to count the gas. Only disable it for debugging, e.g. to obtain a cleaner `opcode.trace`.
* `RuntimeBreakpoints`, default `true`, which causes Wasmer to insert breakpoints into WASM so that Arwen can stop the execution at any time. Only disable it for debugging, e.g. to obtain a cleaner `opcode.trace`.

When `Metering` and `RuntimeBreakpoints` are enabled, the `opcode.trace` is much noisier, because metering- and breakpoint-related code is injected directly into the WASM.

Other changes:
* Create test which executes a WASM file containing `i64.popcnt`; this test fails intentionally fails on older CPUs.
* Minor cleaning in tests.
* Auto-generated files `wasmer.h` and `wasmer.hh` now contain comments from Wasmer's `runtime-c-api` package.


wasmer/wasmer.hh
```diff
@@ -1,17 +1,30 @@
 
 #if !defined(WASMER_H_MACROS)
+
```
andreibancioiu, line 3: Please mention the changes to `wasmer.h` and `wasmer.hh` in the PR. E.g. "comments from upstream".

camilbancioiu: Updated description.

camilbancioiu:
> Include compiled libraries for all supported platforms

It's Github that shows some of them as if they've been deleted. They're included, visible here: https://github.com/ElrondNetwork/arwen-wasm-vm/tree/wasmer-options/wasmer.

PR #136, opened 2020-06-24
Fix canExecuteSynchronously()

camilbancioiu:
Improve the method `host.canExecuteSynchronously()` to allow calls to built-in functions of regular accounts, i.e. accounts that don't hold smart contract code.

### DIFF
--- a/arwen/host/asyncCall.go
+++ b/arwen/host/asyncCall.go
@@ -53,22 +53,34 @@ func (host *vmHost) canExecuteSynchronously(destination []byte, data []byte) boo
 	runtime := host.Runtime()
 	blockchain := host.Blockchain()
 
+	// Sync execution requires the destination to be in the same shard.
 	shardOfSC := blockchain.GetShardOfAddress(runtime.GetSCAddress())
 	shardOfDest := blockchain.GetShardOfAddress(destination)
 	if shardOfSC != shardOfDest {
 		return false
 	}
 
+	// TODO add check for blockchain.IsPayable(destination), when it will be
+	// available.
+
+	// If ArgParser cannot read the Data field, then this isn't even a smart
+	// contract call. Async calls should not be used for transferring value.
 	argParser := parsers.NewCallArgsParser()
 	functionName, _, err := argParser.ParseData(string(data))
 	if err != nil {
 		return false
 	}
 
-	isSmartContract := blockchain.IsSmartContract(destination)
-	isCallToBuiltin := host.isBuiltinFunctionName(functionName)
+	// If the called account has code, it is a smart contract and can be called
+	// without further investigation.
+	code, err := blockchain.GetCode(destination)
+	if len(code) > 0 && err == nil {
+		return true
+	}
 
-	return isSmartContract || isCallToBuiltin
+	// If all else fails, this must be a call to a built-in function to execute
+	// synchronously.
+	return host.isBuiltinFunctionName(functionName)
 }
 
 func (host *vmHost) sendAsyncCallToDestination(asyncCallInfo arwen.AsyncCallInfoHandler) error {
--- a/arwen/host/asyncCall.go
+++ b/arwen/host/asyncCall.go
@@ -18,29 +18,28 @@ func (host *vmHost) handleAsyncCallBreakpoint() error {
 	runtime.SetRuntimeBreakpointValue(arwen.BreakpointNone)
 
 	asyncCallInfo := runtime.GetAsyncCallInfo()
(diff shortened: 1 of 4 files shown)


arwen/host/asyncCall.go
```diff
@@ -51,21 +49,38 @@ func (host *vmHost) handleAsyncCallBreakpoint() error {
...
-	runtime := host.Runtime()
-	asyncCallInfo := runtime.GetAsyncCallInfo()
-	dest := asyncCallInfo.Destination
+	// TODO add check for blockchain.IsPayable(destination), when it will be
+	// available.
+
+	// If ArgParser cannot read the Data field, then this isn't even a smart
+	// contract call. Async calls should not be used for transferring value.
+	argParser := parsers.NewCallArgsParser()
+	functionName, _, err := argParser.ParseData(string(data))
```
sasurobert, line 69: this is not good - instead of canExecuteSynchronously it should be shouldExecuteSynchronously which should return true even if argument parser fails.

arwen/host/asyncCall.go
```diff
@@ -51,21 +49,38 @@ func (host *vmHost) handleAsyncCallBreakpoint() error {
...
+	runtime := host.Runtime()
 	blockchain := host.Blockchain()
-	calledSCCode, err := blockchain.GetCode(dest)
 
-	return len(calledSCCode) != 0 && err == nil
-}
+	// Sync execution requires the destination to be in the same shard.
+	shardOfSC := blockchain.GetShardOfAddress(runtime.GetSCAddress())
+	shardOfDest := blockchain.GetShardOfAddress(destination)
+	if shardOfSC != shardOfDest {
```
sasurobert, line 59: in case of built in functions those has to be processed always in both sender and destination shard as well.

camilbancioiu: Indeed. Will change.

arwen/host/asyncCall.go
camilbancioiu: Will rewrite.

arwen/host/asyncCall.go
```diff
@@ -17,32 +17,32 @@ func (host *vmHost) handleAsyncCallBreakpoint() error {
...
+		}
+		return host.sendAsyncCallToDestination(asyncCallInfo)
 	}
 
-	callbackVMOutput, _, callBackErr := host.ExecuteOnDestContext(callbackCallInput)
+	// Start calling the destination SC, synchronously.
+	destinationVMOutput, destinationErr := host.executeSyncDestinationCall(asyncCallInfo)
+
+	callbackVMOutput, callBackErr := host.executeSyncCallbackCall(asyncCallInfo, destinationVMOutput, destinationErr)
+
```
sasurobert, line 45: delete empty line

arwen/host/asyncCall.go
```diff
@@ -17,32 +17,32 @@ func (host *vmHost) handleAsyncCallBreakpoint() error {
...
+		if err != nil {
+			return err
+		}
+		return host.sendAsyncCallToDestination(asyncCallInfo)
 	}
 
-	callbackVMOutput, _, callBackErr := host.ExecuteOnDestContext(callbackCallInput)
+	// Start calling the destination SC, synchronously.
+	destinationVMOutput, destinationErr := host.executeSyncDestinationCall(asyncCallInfo)
+
```
sasurobert, line 43: delete empty line

arwen/host/asyncCall.go
```diff
@@ -51,21 +51,72 @@ func (host *vmHost) handleAsyncCallBreakpoint() error {
...
 	blockchain := host.Blockchain()
-	calledSCCode, err := blockchain.GetCode(dest)
 
-	return len(calledSCCode) != 0 && err == nil
+	// If ArgParser cannot read the Data field, then this is neither a SC call,
+	// nor a built-in function call.
+	argParser := parsers.NewCallArgsParser()
+	functionName, _, err := argParser.ParseData(string(asyncCallInfo.Data))
+	if err != nil {
+		return arwen.AsyncUnknown, err
```
sasurobert, line 63: nice one !

arwen/host/asyncCall.go
```diff
@@ -51,21 +51,72 @@ func (host *vmHost) handleAsyncCallBreakpoint() error {
...
+	}
+
+	code, err := blockchain.GetCode(asyncCallInfo.Destination)
+	if len(code) > 0 && err == nil {
```
andreibancioiu, line 67: If there's an `err` here we continue, but completely discard the error. Correct?

sasurobert: yes. correct.

PR #137, opened 2020-06-25
return error for empty code

ccorcoveanu:
Return error on `GetCode` if no code is there. This is because on the previous version of BlockchainHook, GetCode was called directly and that function was taking care of that check. Now, using `GetUserAccount` we need to take care of this check our selves. Also separated the error checks to remove the (small) chance of getting a `return nil, nil`

### DIFF
--- a/arwen/errors.go
+++ b/arwen/errors.go
@@ -78,3 +78,5 @@ var ErrShiftNegative = errors.New("bitwise shift operations only allowed on posi
 var ErrAsyncContextDoesNotExist = errors.New("async context was not created yet")
 
 var ErrAsync = errors.New("invalid gas percentage provided for async call")
+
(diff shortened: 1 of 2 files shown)


arwen/errors.go
```diff
@@ -78,3 +78,5 @@ var ErrShiftNegative = errors.New("bitwise shift operations only allowed on posi
 var ErrAsyncContextDoesNotExist = errors.New("async context was not created yet")
 
 var ErrAsync = errors.New("invalid gas percentage provided for async call")
+
+var ErrInvalidAccount = errors.New("account does exist")
```
camilbancioiu, line 82: "account does not exist"

ccorcoveanu: done

PR #147, opened 2020-07-16
Improve instance context management

camilbancioiu:
This PR adds the following changes to the `RuntimeContext` and to the methods in `execution.go`:

* Individual Wasmer instances now contain their own `InstanceContext` object, which obviates the need for maintaining the references separately. This fixes a bug in `elrondei.go`, which caused a panic if the WASM memory was accessed _after_ pushing / popping the instance stack of the `RuntimeContext` (this exact issue was happening in the `createContract()` API method). By putting the `InstanceContext` pointer inside the Wasmer instance, the `InstanceContext` objects are automatically kept on a stack (since Wasmer instances are already stacked).
* Method `RuntimeContext.SetInstanceContextID()` was removed due to the change described above.
* Putting `InstanceContext` objects inside Wasmer instances is the first and most important step for decoupling the `RuntimeContext` from the `vmContextMap`, but removing the `vmContextMap` will come later.
* All calls to `arwen.AddHostContext()` and `arwen.RemoveHostContext()` have been deleted from `execution.go` and moved into the `RuntimeContext`. After the removal of `vmContextMap`, they will be removed as well, eventually.

### DIFF
--- a/arwen/contexts/runtime.go
+++ b/arwen/contexts/runtime.go
@@ -289,14 +289,6 @@ func (context *runtimeContext) SetReadOnly(readOnly bool) {
 	context.readOnly = readOnly
 }
 
-func (context *runtimeContext) SetInstanceContext(instCtx *wasmer.InstanceContext) {
-	// TODO remove this method when the vmContextMap is removed
-}
-
-func (context *runtimeContext) GetInstanceContext() *wasmer.InstanceContext {
-	return &context.instance.InstanceCtx
-}
-
 func (context *runtimeContext) GetInstanceExports() wasmer.ExportsMap {
 	return context.instance.Exports
 }
--- a/arwen/contexts/runtime.go
+++ b/arwen/contexts/runtime.go
@@ -11,14 +11,13 @@ import (
 var _ arwen.RuntimeContext = (*runtimeContext)(nil)
 
 type runtimeContext struct {
-	host            arwen.VMHost
-	instance        *wasmer.Instance
-	instanceContext *wasmer.InstanceContext
-	vmInput         *vmcommon.VMInput
-	scAddress       []byte
-	callFunction    string
-	vmType          []byte
-	readOnly        bool
+	host         arwen.VMHost
+	instance     *wasmer.Instance
+	vmInput      *vmcommon.VMInput
+	scAddress    []byte
+	callFunction string
+	vmType       []byte
+	readOnly     bool
 
 	stateStack    []*runtimeContext
 	instanceStack []*wasmer.Instance
@@ -291,11 +290,11 @@ func (context *runtimeContext) SetReadOnly(readOnly bool) {
 }
 
 func (context *runtimeContext) SetInstanceContext(instCtx *wasmer.InstanceContext) {
-	context.instanceContext = instCtx
+	// TODO remove this method when the vmContextMap is removed
 }
 
 func (context *runtimeContext) GetInstanceContext() *wasmer.InstanceContext {
-	return context.instanceContext
+	return &context.instance.InstanceCtx
 }
 
 func (context *runtimeContext) GetInstanceExports() wasmer.ExportsMap {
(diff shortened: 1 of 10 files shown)


arwen/contexts/runtime.go
```diff
@@ -83,6 +77,16 @@ func (context *runtimeContext) StartWasmerInstance(contract []byte, gasLimit uin
 	}
 
 	context.instance = newInstance
+
+	idContext := arwen.AddHostContext(context.host)
+	context.instance.SetContextData(idContext)
+
+	err = context.VerifyContractCode()
```
andreibancioiu, line 84: So we verify contract code again here, as well - not only at deploy time as before?

camilbancioiu: Yes, I added this as a defense-in-depth measure. Arwen shouldn't immediately trust the code sent by the node, even if it was already deployed.

`VerifyContractCode()` isn't doing that many verifications, and it should be extended in the future with extra ones.

sasurobert: I do not think this verification is needed. There is no things like malicious elrond-go or malicious arwen. We should not treat them as sending wrong data from one to another.

arwen/contexts/runtime.go
```diff
@@ -285,27 +289,24 @@ func (context *runtimeContext) SetReadOnly(readOnly bool) {
...
 }
 
-func (context *runtimeContext) SetInstanceContextID(id int) {
-	context.instanceContextDataPointers = append(context.instanceContextDataPointers, &id)
-	context.instance.SetContextData(unsafe.Pointer(&id))
-}
-
 func (context *runtimeContext) SetInstanceContext(instCtx *wasmer.InstanceContext) {
-	context.instanceContext = instCtx
+	// TODO remove this method when the vmContextMap is removed
```
sasurobert, line 293: this function does not do anything. you can safely remove it.

wasmer/import.go
```diff
@@ -199,6 +199,12 @@ func IntoInstanceContext(instanceContext unsafe.Pointer) InstanceContext {
 	return InstanceContext{context, memory}
 }
 
+func IntoInstanceContextDirect(instanceContext *cWasmerInstanceContextT) InstanceContext {
+	memory := newMemory(cWasmerInstanceContextMemory(instanceContext))
+
```
sasurobert, line 204: delete empty line

arwen/contexts/runtime.go
camilbancioiu: Removed, along with `GetInstanceContext()`.

arwen/contexts/runtime.go
camilbancioiu: `VerifyContractCode()` is needed for SC deployment, but it's almost useless for SC calls. Still, having it here simplifies the code tremendously in `execution.go`. As a side-effect, `VerifyContractCode()` gets called for every single Wasmer instance because switching it on for the SC deployment and then switching it off for SC call was making the code terrible. I tried. This was the simplest solution.

The true issue: `VerifyContractCode()` is useless for SC calls, but we must improve that (any ideas?), and use it whenever starting a Wasmer instance. 

The node is complex and valuable enough to make hacking it [(trivial example)](https://codearcana.com/posts/2013/04/23/exploiting-a-go-binary.html) a definite possibility, then exploiting its own code against itself, [without an external payload](https://en.wikipedia.org/wiki/Return-oriented_programming). And the OS can't help, therefore we need as many security checks in general, everywhere (especially in `elrond-go`; less in Arwen). Calling `VerifyContractCode()` everytime seems pointless now because it doesn't contain enough checks.

Golang does have many checks itself, but [it's easy to mistakenly write insecure code in Golang](https://sca.analysiscenter.veracode.com/vulnerability-database/search#query=language:go), just like in other languages.

The other important issue: the WASM bytecode is traversing an environmental boundary. Not Protocol → Arwen, but "storage" → "execution". For example, web apps have a lot of sanitization before displaying anything stored in the database. Same here, assuming `VerifyContractCode()` does something meaningful for SC calls (not yet the case).

sasurobert: I still think there is no need to verifyContractCode for every instance. The CODE which gets deployed is always checked - verified. There is no possibility in case of how elrond is built for arwen to receive a code which was not checked.

camilbancioiu: Added an extra option to the `RuntimeContext` to enable verification only for deployment, but in the next PR. This PR should be merged as-is for now.

PR #149, opened 2020-07-21
Check if contract is payable in Transfer function

ccorcoveanu:
- `go.mod` and `go.sum`will be updated after the vmcommon PR is merged and a new release is available
 - Had to make a decision between having a new hook for this functionality or load the account and make the same checks as in `elrond-go`. I went with this options so we won't have to pass the whole account between processes, but this can be changed


arwen/contexts/output.go
```diff
@@ -199,6 +199,16 @@ func (context *outputContext) Transfer(destination []byte, sender []byte, gasLim
...
+	payable, err := context.host.Blockchain().IsPayable(destination)
+	if err != nil {
+		return err
+	}
+
+	hasValue := value.Cmp(big.NewInt(0)) == 1
+	if !payable && hasValue {
+		return arwen.ErrAccountNotPayable
+	}
+
```
camilbancioiu, line 211: Move these lines above `senderAcc := ...`, to return earlier in case of error.

arwen/contexts/output_test.go
```diff
@@ -460,6 +460,45 @@ func TestOutputContext_Transfer_Errors_And_Checks(t *testing.T) {
 	require.Equal(t, big.NewInt(1000), blockchainContext.GetBalanceBigInt(sender))
 }
 
+func TestOutputContext_Transfer_IsAccountPayable(t *testing.T) {
+	t.Parallel()
+
+	sender := []byte("sender")
+	receiverNonPayable := make([]byte, 32)
```
camilbancioiu, line 467: Also add a new SC address to the test, `receiverPayable`, and add another call to `outputContext.Transfer()` from `sender` to `receiverPayable` which should work.

ipc/arwenpart/blockchainGateway.go
```diff
@@ -351,3 +353,21 @@ func (blockchain *BlockchainHookGateway) IsSmartContract(address []byte) bool {
...
+func (blockchain *BlockchainHookGateway) IsPayable(address []byte) (bool, error) {
+	request := common.NewMessageBlockchainIsPayableRequest(address)
+	rawResponse, err := blockchain.messenger.SendHookCallRequest(request)
+	if err != nil {
+		return false, err
+	}
+
+	if rawResponse.GetKind() != common.BlockchainIsPayableResponse {
+		x := rawResponse.GetKind()
+		fmt.Println(x)
```
camilbancioiu, line 367: Remove `fmt.Println()`.

arwen/contexts/output.go
ccorcoveanu: done

arwen/contexts/output_test.go
ccorcoveanu: done

ipc/arwenpart/blockchainGateway.go
ccorcoveanu: done

arwen/contexts/output.go
```diff
@@ -196,6 +196,16 @@ func (context *outputContext) Transfer(destination []byte, sender []byte, gasLim
 		return arwen.ErrTransferInsufficientFunds
 	}
 
+	payable, err := context.host.Blockchain().IsPayable(destination)
+	if err != nil {
+		return err
+	}
+
+	hasValue := value.Cmp(big.NewInt(0)) == 1
```
andreibancioiu, line 204: It would be nice to have some functions in `helpers.go` such as `bigIntGreaterThan()` - not for now.

ipc/arwenpart/blockchainGateway_test.go
```diff
@@ -118,6 +118,21 @@ func TestGateway_IsSmartContract(t *testing.T) {
 	runHookScenario(t, callHook, handleHookCall)
 }
 
+func TestGateway_IsPayable(t *testing.T) {
```
andreibancioiu, line 121: :+1: for also testing this.

PR #151, opened 2020-07-24
Wrap VMOutput in a JSON-friendly structure

andreibancioiu:
Issue:

 - `VMOutput` contains nested `map[string]interface{}`, and this leads to have the map keys UTF-8 encoded in the context of JSON serialization (required by Node - Arwen Inter Process Communication).
 - Though, the keys could contain JSON-invalid (non UTF-8) sequences of bytes
 - Such keys were subject to UTF-8 encoding replacements - which could sometimes result in undesired de-duplication of map keys at deserialization time.

Changes:
 - Wrapped `VMOutput` in a JSON-friendly structure
 - Added tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/ElrondNetwork/elrond-go-logger v1.0.2
 	github.com/ElrondNetwork/elrond-vm-common v0.1.23
 	github.com/ElrondNetwork/elrond-vm-util v0.3.5
+	github.com/davecgh/go-spew v1.1.1
 	github.com/gin-gonic/gin v1.6.3
 	github.com/mitchellh/mapstructure v1.1.2
 	github.com/pelletier/go-toml v1.6.0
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/ElrondNetwork/elrond-go-logger v1.0.2
 	github.com/ElrondNetwork/elrond-vm-common v0.1.23
 	github.com/ElrondNetwork/elrond-vm-util v0.3.5
+	github.com/davecgh/go-spew v1.1.1
```
sasurobert, line 10: maybe delete this

andreibancioiu: Fixed.

ipc/common/messagesBlockchain.go
```diff
@@ -469,6 +469,7 @@ func NewMessageBlockchainGetAllStateRequest(address []byte) *MessageBlockchainGe
 // MessageBlockchainGetAllStateResponse represents a response message
 type MessageBlockchainGetAllStateResponse struct {
 	Message
+	// TODO: Make sure to change "map[string][]byte" to a JSON-serializable friendly structure when "GetAllState()" will be used
```
camilbancioiu, line 472: There should be a JIRA task for this, if there's not one open already.

PR #152, opened 2020-07-24
More minor improvements

camilbancioiu:
This PR contains the following changes:
* `outputContext.GetStorage()` and `outputContext.SetStorage()` now consume gas based on the length of the provided key, if the key is longer than 32 bytes.
* Calling `outputContext.GetStorage()` now consumes gas directly, so that the API functions that call it do not have to calculate the length of the data loaded from storage themselves.
* Added the function `outputContext.GetStorageUnmetered()`, which retrieves data from storage without metering. It is used as a utility by other methods, but it is never called directly by API functions that are called by smart contracts, except in `storageLoadLength()`.
* Added proper protection against overwriting existing SCs when using `createContract()`. Previous implementation provided only superficial and incomplete protection.
* Added protection against calling any built-in function using `transferValue()`, using an argument parser.
* Added a clean-up phase when generating the `VMOutput`, which removes from `OutputAccounts` all the `.Code` fields that were not actually deployed or upgraded. This prevents the unchanged code of smart contracts from being propagated back to the node via the `VMOutput`, just because the code was retrieved at some point during execution.
* Many API functions in `elrondei.go` and `bigintOps.go` have been changed to consume gas before returning in case of error. This prevents abusive repeated calls to expensive API functions which end up failing, but without any gas consumption.

TODO: 
* tests for gas consumption for keys longer than 32 bytes
* tests for protecting against calls to built-in functions with `transferValue()`

### DIFF
--- a/arwen/contexts/storage.go
+++ b/arwen/contexts/storage.go
@@ -71,16 +71,35 @@ func (context *storageContext) GetStorageUpdates(address []byte) map[string]*vmc
 }
 
 func (context *storageContext) GetStorage(key []byte) []byte {
-	storageUpdates := context.GetStorageUpdates(context.address)
-	if storageUpdate, ok := storageUpdates[string(key)]; ok {
-		return storageUpdate.Data
+	metering := context.host.Metering()
+
+	extraBytes := len(key) - arwen.AddressLen
+	if extraBytes > 0 {
+		gasToUse := metering.GasSchedule().BaseOperationCost.DataCopyPerByte * uint64(extraBytes)
+		metering.UseGas(gasToUse)
 	}
 
-	value, _ := context.blockChainHook.GetStorageData(context.address, key)
-	if value != nil {
-		storageUpdates[string(key)] = &vmcommon.StorageUpdate{
-			Offset: key,
-			Data:   value,
+	value := context.GetStorageUnmetered(key)
+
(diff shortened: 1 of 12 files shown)


arwen/elrondapi/bigintOps.go
```diff
@@ -391,12 +390,12 @@ func bigIntSignedByteLength(context unsafe.Pointer, reference int32) int32 {
...
 	metering := arwen.GetMeteringContext(context)
 
-	value := bigInt.GetOne(reference)
-
 	gasToUse := metering.GasSchedule().BigIntAPICost.BigIntSignedByteLength
 	metering.UseGas(gasToUse)
 
-	bytes := twos.ToBytes(value) // TODO: figure out the correct length without computing the 2's complement
+	value := bigInt.GetOne(reference)
+
```
sasurobert, line 397: Delete empty line

arwen/contexts/storage.go
```diff
@@ -71,16 +71,35 @@ func (context *storageContext) GetStorageUpdates(address []byte) map[string]*vmc
...
-			Data:   value,
+	value := context.GetStorageUnmetered(key)
+
+	gasToUse := metering.GasSchedule().BaseOperationCost.DataCopyPerByte * uint64(len(value))
+	metering.UseGas(gasToUse)
+
+	return value
+}
+
+func (context *storageContext) GetStorageUnmetered(key []byte) []byte {
```
sasurobert, line 90: +1

PR #153, opened 2020-07-26
no need to ask user storage multiple times for empty value

sasurobert:


arwen/host/execution_test.go
```diff
@@ -346,10 +346,10 @@ func TestExecution_ExecuteOnSameContext_Simple(t *testing.T) {
 	input.GasProvided = gasProvided
 
 	vmOutput, err := host.RunSmartContractCall(input)
-	fmt.Println(vmOutput.ReturnMessage)
 	require.Nil(t, err)
 	require.Equal(t, vmcommon.Ok, vmOutput.ReturnCode)
 	require.Equal(t, "", vmOutput.ReturnMessage)
+	fmt.Println(vmOutput.ReturnMessage)
```
andrei-marinica, line 352: maybe delete this altogether?

sasurobert: left it as it is, maybe needed for debug

iulianpascalau: use log then

PR #155, opened 2020-07-27
Fix serialization for IPC messages

andreibancioiu:
Fix serialization for IPC messages:
 - built-in function calls (vm output)
 - `getAllState`

### DIFF
--- a/arwen/contexts/output.go
+++ b/arwen/contexts/output.go
@@ -328,8 +328,8 @@ func (context *outputContext) AddToActiveState(rightOutput *vmcommon.VMOutput) {
 		rightOutput.GasRefund.Add(rightOutput.GasRefund, context.outputState.GasRefund)
 	}
 
-	for _, rightAccount := range rightOutput.OutputAccounts {
-		leftAccount, ok := context.outputState.OutputAccounts[string(rightAccount.Address)]
+	for address, rightAccount := range rightOutput.OutputAccounts {
+		leftAccount, ok := context.outputState.OutputAccounts[address]
 		if !ok || rightAccount.BalanceDelta == nil {
 			continue
 		}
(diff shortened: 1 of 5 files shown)


arwen/contexts/output.go
```diff
@@ -328,8 +328,8 @@ func (context *outputContext) AddToActiveState(rightOutput *vmcommon.VMOutput) {
 		rightOutput.GasRefund.Add(rightOutput.GasRefund, context.outputState.GasRefund)
 	}
 
-	for _, rightAccount := range rightOutput.OutputAccounts {
-		leftAccount, ok := context.outputState.OutputAccounts[string(rightAccount.Address)]
+	for address, rightAccount := range rightOutput.OutputAccounts {
```
sasurobert, line 331: please move it back :) it is much safer.

arwen/contexts/output.go
```diff
@@ -345,11 +345,12 @@ func mergeVMOutputs(leftOutput *vmcommon.VMOutput, rightOutput *vmcommon.VMOutpu
 	if leftOutput.OutputAccounts == nil {
 		leftOutput.OutputAccounts = make(map[string]*vmcommon.OutputAccount)
 	}
-	for _, rightAccount := range rightOutput.OutputAccounts {
-		leftAccount, ok := leftOutput.OutputAccounts[string(rightAccount.Address)]
+
+	for address, rightAccount := range rightOutput.OutputAccounts {
```
sasurobert, line 349: please move it back :) it is much safer.

ipc/common/messagesBlockchain.go
```diff
@@ -412,14 +412,14 @@ func NewMessageBlockchainProcessBuiltinFunctionRequest(callInput vmcommon.Contra
...
-	VMOutput *vmcommon.VMOutput
+	SerializableVMOutput *SerializableVMOutput
 }
 
 // NewMessageBlockchainProcessBuiltinFunctionResponse creates a response message
 func NewMessageBlockchainProcessBuiltinFunctionResponse(vmOutput *vmcommon.VMOutput, err error) *MessageBlockchainProcessBuiltinFunctionResponse {
 	message := &MessageBlockchainProcessBuiltinFunctionResponse{}
 	message.Kind = BlockchainProcessBuiltinFunctionResponse
-	message.VMOutput = vmOutput
+	message.SerializableVMOutput = NewSerializableVMOutput(vmOutput)
```
sasurobert, line 422: add a unit test ?

arwen/contexts/output.go
andreibancioiu: Reverted.

ipc/common/messagesBlockchain.go
andreibancioiu: Added.

arwen/contexts/output.go
andreibancioiu: Reverted.

arwen/contexts/output.go
camilbancioiu: Using `string(rightAccount.Address)` will still cause problems if there are duplicate `OutputAccounts`, due to random iteration over maps. It doesn't actually help, it only hides the problem. The duplication was the issue, not this iteration.

arwen/contexts/output.go
camilbancioiu: See above.

PR #156, opened 2020-07-28
Contract test 7

andrei-marinica:
`transferValue` fails if value exceeds balance.
Some updated tests.


arwen/elrondapi/elrondei.go
```diff
@@ -512,7 +520,7 @@ func transferValue(context unsafe.Pointer, destOffset int32, valueOffset int32,
 		return 1
 	}
 
-	err = output.Transfer(dest, send, 0, big.NewInt(0).SetBytes(value), data)
+	err = output.Transfer(dest, send, 0, value, data)
 	if err != nil {
```
camilbancioiu, line 524: If you want `transferValue()` to fail the entire execution in case of error, change this to `if arwen.WithFault(err, context, runtime.ElrondAPIErrorShouldFailExecution()) {`. Otherwise, use the `return` value to determine success / failure of the transfer.

PR #170, opened 2020-09-18
Read data from another contract

sasurobert:
implemented reading from another address / smart contract if that isreadable


arwen/elrondapi/elrondei.go
```diff
@@ -776,6 +782,35 @@ func storageLoadLength(context unsafe.Pointer, keyOffset int32, keyLength int32)
 	return int32(len(data))
 }
 
+//export storageLoadFromAddress
+func storageLoadFromAddress(context unsafe.Pointer, addressOffset int32, keyOffset int32, keyLength int32, dataOffset int32) int32 {
+	runtime := arwen.GetRuntimeContext(context)
+	storage := arwen.GetStorageContext(context)
+	metering := arwen.GetMeteringContext(context)
+
+	gasToUse := metering.GasSchedule().ElrondAPICost.StorageLoad
```
andreibancioiu, line 791: Just to double check - we don't need to define another cost for this, right?

andreibancioiu (approved):

arwen/elrondapi/elrondei.go
sasurobert: no need. same cost

ccorcoveanu (approved):

PR #180, opened 2020-10-21
Small Int EI

andrei-marinica:
Small int api, properly named and implemented.

Old functions kept for backwards compatibility.

Some more validation, tests.
The framework can disable the small int API, so there's a test with that too.


arwen/elrondapi/elrondei.go
```diff
@@ -1359,9 +1384,13 @@ func int64storageLoad(context unsafe.Pointer, keyOffset int32, keyLength int32)
 
 	data := storage.GetStorage(key)
 
-	bigInt := big.NewInt(0).SetBytes(data)
+	valueBigInt := twos.SetBytes(big.NewInt(0), data)
```
sasurobert, line 1387: is this backward compatible ?

camilbancioiu: Do we need the same in `int64storageStore()`?

andrei-marinica: no :(
will redesign

andrei-marinica: @camil yes, slipped me.
But will rewrite anyway

camilbancioiu (approved):

sasurobert (approved):

PR #188, opened 2020-11-04
Debug arwen oop

sasurobert:


camilbancioiu (dismissed):

andrei-marinica (approved):

andrei-marinica (approved):

ipc/nodepart/arwenDriver.go
```diff
@@ -109,6 +109,8 @@ func (driver *ArwenDriver) startArwen() error {
 		return err
 	}
 
+	driver.blockchainHook.ClearCompiledCodes()
```
andreibancioiu, line 112: Thus we have to make sure we never start the querying engine (at some random time) using OOP (Arwen Driver). Because that will clear the cache as well. Should never happen in practice, actually.

andreibancioiu (approved):

ipc/nodepart/arwenDriver.go
andreibancioiu: If the cache is shared (I think that's the case).

sasurobert: yes. that is the case. SO ScQuery and normal processing - with different setups will not work. I will make a different cache for scQuery in elrond go. good comment.

PR #189, opened 2020-11-04
Fix execution tests

camilbancioiu:
* Changed execution tests to do the correct asserts for setting `ElrondSyncExecAPIErrorShouldFailExecution` either true or false.
* Fixed `TestExecution_ExecuteOnSameContext_Simple`, which was broken.


arwen/contexts/runtime.go
```diff
@@ -461,6 +461,10 @@ func (context *runtimeContext) ElrondAPIErrorShouldFailExecution() bool {
 	return true
 }
 
+func (context *runtimeContext) ElrondSyncExecAPIErrorShouldFailExecution() bool {
+	return false
```
sasurobert, line 465: this should not be false.

arwen/elrondapi/elrondei.go
```diff
@@ -1291,18 +1291,18 @@ func executeOnSameContext(
 
 	sender := runtime.GetSCAddress()
 	destination, err := runtime.MemLoad(addressOffset, arwen.AddressLen)
-	if arwen.WithFault(err, context, runtime.ElrondAPIErrorShouldFailExecution()) {
+	if arwen.WithFault(err, context, runtime.ElrondSyncExecAPIErrorShouldFailExecution()) {
```
sasurobert, line 1294: no need for this

sasurobert: make it back

sasurobert (changes requested):

arwen/host/execution_test.go
```diff
@@ -772,8 +797,16 @@ func TestExecution_ExecuteOnDestContext_Wrong(t *testing.T) {
 
 	vmOutput, err := host.RunSmartContractCall(input)
 	require.Nil(t, err)
-	expectedVMOutput := expectedVMOutput_DestCtx_WrongContractCalled(parentCode)
-	require.Equal(t, expectedVMOutput, vmOutput)
+	require.NotNil(t, vmOutput)
+
+	if host.Runtime().ElrondSyncExecAPIErrorShouldFailExecution() == false {
```
andrei-marinica, line 802: weird having == false, maybe make it !host..... or invert the conditions

andrei-marinica (approved):

arwen/host/execution_test.go
sasurobert: this is only to check that the code was there.

PR #214, opened 2020-12-09
EN-8423-safe-uint-math

BeniaminDrasovean:


go.sum
```diff
@@ -522,6 +522,7 @@ github.com/onsi/gomega v1.9.0 h1:R1uwffexN6Pr340GtYRIdZmAiN4J+iw6WG4wog1DUXg=
 github.com/onsi/gomega v1.9.0/go.mod h1:Ho0h+IUsWyvy1OpqCwxlQ/21gkhVunqlU8fDGcoTdcA=
 github.com/opentracing/opentracing-go v1.0.2/go.mod h1:UkNAQd3GIcIGf0SeVgPpRdFStlNbqXla1AfSYxPUl2o=
 github.com/opentracing/opentracing-go v1.1.0/go.mod h1:UkNAQd3GIcIGf0SeVgPpRdFStlNbqXla1AfSYxPUl2o=
+github.com/opentracing/opentracing-go v1.2.0 h1:uEJPy/1a5RIPAJ0Ov+OIO8OxWu77jEv+1B0VhjKrZUs=
```
sasurobert, line 525: go mod tidy

math/overflow.go
```diff
@@ -1,11 +1,94 @@
...
+	if (res > a) == (b > 0) {
+		return res
+	}
+
+	log.Error("AddInt64 overflow", a, b)
+	return builtinMath.MaxInt64
+}
+
+// SubUint64 performs subtraction on uint64 and logs an error if the subtraction overflows
+func SubUint64(a, b uint64) uint64 {
```
sasurobert, line 44: maybe keep the same standard as with Add - where you have SubUnitWithError as well.

sasurobert (changes requested):

arwen/contexts/metering.go
```diff
@@ -50,8 +51,8 @@ func (context *meteringContext) SetGasSchedule(gasMap config.GasScheduleMap) {
 
 // UseGas sets in the runtime context the given gas as gas used
 func (context *meteringContext) UseGas(gas uint64) {
-	gasUsed := context.host.Runtime().GetPointsUsed() + gas
-	context.host.Runtime().SetPointsUsed(gasUsed)
+	gasUsed := math.AddInt64(int64(context.host.Runtime().GetPointsUsed()), int64(gas))
```
camilbancioiu, line 54: Create a variable `gas_i64` and set it to `MaxInt64` (signed) in case the `gas` argument (unsigned) would overflow during the conversion `int64(gas)`. Otherwise, a contract that somehow managed to force a situation with `metering.UseGas(MaxInt64 + 10)` would actually regain `9` gas due to the u64 → i64 overflow. 

This is a hypothetical issue, and is not expected to occur in practice.

camilbancioiu: Same would go for `context.host.Runtime().GetPointsUsed()` - it's the same u64 → i64 conversion. 

However, these conversions should probably be handled in the Runtime context, in the methods `GetPointsUsed()` and `SetPointsUsed()`. Can you please add conditions in these two methods, to ensure that they stay within the bounds of i64? It's safest.

arwen/contexts/metering.go
```diff
@@ -194,8 +198,25 @@ func (context *meteringContext) deductInitialGas(
 ) error {
 	input := context.host.Runtime().GetVMInput()
 	codeLength := uint64(len(code))
-	codeCost := codeLength * costPerByte
-	initialCost := baseCost + codeCost
+	codeCost, err := math.MulUint64WithErr(codeLength, costPerByte)
```
camilbancioiu, line 201: I think it's safe to use `math.MulUint64()` directly, without `Err`, making the code cleaner. The condition at line 221 would prevent using `MaxUint64` gas.

arwen/contexts/metering.go
```diff
@@ -194,8 +198,25 @@ func (context *meteringContext) deductInitialGas(
...
+	if err != nil {
+		log.Error("deductInitialGas overflow",
+			"codeLength = ", codeLength,
+			"costPerByte = ", costPerByte,
+		)
+
+		return arwen.ErrNotEnoughGas
+	}
+
+	initialCost, err := math.AddUint64WithErr(baseCost, codeCost)
```
camilbancioiu, line 211: Same as above, `math.AddUint64()` without `Err` would be cleaner, yet equivalent due to the condition at 221.

arwen/contexts/runtime.go
```diff
@@ -688,10 +689,14 @@ func (context *runtimeContext) MemLoad(offset int32, length int32) ([]byte, erro
 	memory := context.instance.InstanceCtx.Memory()
 	memoryView := memory.Data()
 	memoryLength := memory.Length()
-	requestedEnd := uint32(offset + length)
+	requestedEnd, err := math.AddInt32(offset, length)
```
camilbancioiu, line 692: Use a variant of `AddInt32` which simply returns MaxInt32 if the addition overflows, and no error. The condition at 702 makes it safe.

camilbancioiu (changes requested):

go.sum
BeniaminDrasovean: Done.

math/overflow.go
BeniaminDrasovean: Done.

math/errors.go
```diff
@@ -4,4 +4,11 @@ import (
 	"errors"
 )
 
+// ErrAdditionOverflow is raised when there is an overflow because of the addition of two numbers
 var ErrAdditionOverflow = errors.New("addition overflow")
+
+// ErrSubtractionOverflow is raised when there is an overflow because of the subtraction of two numbers
+var ErrSubtractionOverflow = errors.New("subtraction overflow")
```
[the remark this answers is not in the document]
BeniaminDrasovean: Done.

math/overflow.go
```diff
@@ -1,11 +1,94 @@
...
 		return s, nil
 	}
 
 	return s, ErrAdditionOverflow
 }
+
+// AddInt64 performs addition on int64 and logs an error if the addition overflows
+func AddInt64(a, b int64) int64 {
+	res := a + b
+	if (res > a) == (b > 0) {
```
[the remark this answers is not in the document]
BeniaminDrasovean: Yes. Both conditions have to be either true or false.

arwen/contexts/metering.go
BeniaminDrasovean: Added protection for SetPointsUsed(), and used Uint64 operations everywhere else.

arwen/contexts/metering.go
BeniaminDrasovean: Done.

arwen/contexts/metering.go
BeniaminDrasovean: Done.

arwen/contexts/runtime.go
BeniaminDrasovean: Done.

camilbancioiu (dismissed):

sasurobert (dismissed):

camilbancioiu (approved):

sasurobert (approved):

PR #213, opened 2020-12-08
Create executeOnDestContextByCaller

camilbancioiu:
Create EEI method `executeOnDestContextByCaller()`, and simplify the methods `executeOnDestContext()`, `executeOnSameContext()`, `delegateExecution()` and `executeReadOnly()`.

PR #215, opened 2020-12-09
Fix Runtime instance stacking

camilbancioiu:
This PR unifies the state stack and instance stack of the `RuntimeContext`, with the following consequences:
* Code is simplified, because the primitive method `executeSmartContractCall()` will neither push nor pop the Runtime instance stack; this means that the code that calls `executeSmartContractCall()` must handle both state stacking and the remaining gas, after the contract is executed indirectly.
* Fixes a gas-related issue on `ExecuteOnSameContext()` and `ExecuteOnDestContext()`, where `GetPointsUsed()` was called on the wrong Wasmer instance.


arwen/elrondapi/elrondei.go
```diff
@@ -1732,10 +1647,105 @@ func getOriginalTxHash(context unsafe.Pointer, dataOffset int32) {
...
+			return "", nil, 0, err
+		}
+
+		currOffset += actualLen
+	}
+
+	return string(function), data, currOffset - dataOffset, nil
+}
+
+func bytesToInt32(data []byte) int32 {
```
andrei-marinica, line 1744: pretty sure this function can be taken from some library
I think this one does it:
https://golang.org/pkg/encoding/binary/#example_ByteOrder_get

andrei-marinica (changes requested):

arwen/elrondapi/elrondei.go
```diff
@@ -1732,10 +1647,105 @@ func getOriginalTxHash(context unsafe.Pointer, dataOffset int32) {
 	_ = arwen.WithFault(err, context, runtime.ElrondAPIErrorShouldFailExecution())
 }
 
-func areInSameShard(host arwen.VMHost, leftAddress []byte, rightAddress []byte) bool {
-	blockchain := host.Blockchain()
-	leftShard := blockchain.GetShardOfAddress(leftAddress)
-	rightShard := blockchain.GetShardOfAddress(rightAddress)
+func prepareIndirectContractCallInput(
```
sasurobert, line 1650: this PR has some changes from the previous one which is not yet merged. let's do them step by step.

sasurobert (changes requested):

arwen/elrondapi/elrondei.go
camilbancioiu: Indeed, this depends on https://github.com/ElrondNetwork/arwen-wasm-vm/pull/213 must be merged first.

arwen/elrondapi/elrondei.go
camilbancioiu: Changed to use `binary.LittleEndian.Uint32()`.

andrei-marinica (approved):

sasurobert (approved):

PR #218, opened 2020-12-13
Implement storage protection

camilbancioiu:
This PR defines the key prefix `ARWEN@` which will be write-protected by the `StorageContext`. 

However, the VM may allow writing to protected keys using the new methods `storage.DisableStorageProtection()` and `storage.EnableStorageProtection()`. This is required for the Promises API and the Timelocks API, which must write to protected storage.

This form of storage protection is only enforced in the VM. The Elrond node always allows `StorageUpdates` to the keys that are otherwise protected by Arwen.

### DIFF
--- a/arwen/contexts/storage.go
+++ b/arwen/contexts/storage.go
@@ -10,11 +10,12 @@ import (
 )
 
 type storageContext struct {
-	host                     arwen.VMHost
-	blockChainHook           vmcommon.BlockchainHook
-	address                  []byte
-	stateStack               [][]byte
-	elrondProtectedKeyPrefix []byte
+	host                          arwen.VMHost
+	blockChainHook                vmcommon.BlockchainHook
+	address                       []byte
+	stateStack                    [][]byte
+	elrondProtectedKeyPrefix      []byte
+	arwenStorageProtectionEnabled bool
 }
 
 // NewStorageContext creates a new storageContext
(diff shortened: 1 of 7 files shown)


arwen/elrondapi/elrondei.go
```diff
@@ -935,7 +935,9 @@ func setStorageLock(context unsafe.Pointer, keyOffset int32, keyLength int32, lo
 
 	timeLockKey := arwen.CustomStorageKey(arwen.TimeLockKeyPrefix, key)
 	bigTimestamp := big.NewInt(0).SetInt64(lockTimestamp)
+	storage.DisableStorageProtection()
 	storageStatus, err := storage.SetStorage(timeLockKey, bigTimestamp.Bytes())
```
sasurobert, line 939: why not make a separate function which does SetStorageToProtectedField ?

sasurobert (changes requested):

sasurobert (approved):

arwen/contexts/storage.go
```diff
@@ -154,18 +156,41 @@ func (context *storageContext) GetStorageUnmetered(key []byte) []byte {
...
+
+func (context *storageContext) isArwenProtectedKey(key []byte) bool {
+	return bytes.HasPrefix(key, []byte(arwen.ProtectedStoragePrefix))
+}
+
 func (context *storageContext) isElrondReservedKey(key []byte) bool {
 	return bytes.HasPrefix(key, context.elrondProtectedKeyPrefix)
 }
 
+func (context *storageContext) SetProtectedStorage(key []byte, value []byte) (arwen.StorageStatus, error) {
```
BeniaminDrasovean, line 177: Missing comment.

BeniaminDrasovean (approved):

arwen/contexts/storage.go
camilbancioiu: Will add on a different branch, thank you.

PR #222, opened 2020-12-15
Compute GasUsed in the MeteringContext

camilbancioiu:
This PR does the following:
* Introduces the concept of "forwarded gas", which is gas spent by a contract for the execution of **other** contracts, which means `executeOnSame/DestContext()` and `asyncCall()`; the amount of "forwarded gas" is deducted from the total gas metered for a contract, to ensure that `outputAccount.GasUsed` contains the amount used for compilation+execution exclusively, and not for other contracts, thus isolating the gas of a contract from the gas consumed by the contract for other contracts;
* Adds state stacking to the `MeteringContext`; each state item stores the initial gas provided and the gas forwarded by a contract;
* Does minor rewriting to `host.execute()`
* Adds a few tests

TODO computing `GasUsed` should have a lot more tests, but the existing ones are passing. Gas invariant checks will confirm the correctness, though.

### DIFF
--- a/arwen/contexts/metering.go
+++ b/arwen/contexts/metering.go
@@ -8,9 +8,12 @@ import (
 )
 
 type meteringContext struct {
-	gasSchedule   *config.GasCost
-	blockGasLimit uint64
-	host          arwen.VMHost
+	host               arwen.VMHost
+	stateStack         []*meteringContext
+	gasSchedule        *config.GasCost
+	blockGasLimit      uint64
+	initialGasProvided uint64
+	gasForwarded       uint64
 }
 
 // NewMeteringContext creates a new meteringContext
@@ -26,12 +29,72 @@ func NewMeteringContext(
 	}
 
 	context := &meteringContext{
+		host:          host,
+		stateStack:    make([]*meteringContext, 0),
 		gasSchedule:   gasSchedule,
 		blockGasLimit: blockGasLimit,
-		host:          host,
 	}
 
+	context.InitState()
+
 	return context, nil
+}
+
+func (context *meteringContext) InitState() {
+	context.initialGasProvided = 0
+	context.gasForwarded = 0
+}
+
+func (context *meteringContext) PushState() {
+	newState := &meteringContext{
+		initialGasProvided: context.initialGasProvided,
+		gasForwarded:       context.gasForwarded,
+	}
+
+	context.stateStack = append(context.stateStack, newState)
+}
+
+func (context *meteringContext) PopSetActiveState() {
+	stateStackLen := len(context.stateStack)
+	if stateStackLen == 0 {
+		return
+	}
+
+	prevState := context.stateStack[stateStackLen-1]
+	context.stateStack = context.stateStack[:stateStackLen-1]
+
+	context.initialGasProvided = prevState.initialGasProvided
+	context.gasForwarded = prevState.gasForwarded
+}
+
+func (context *meteringContext) PopDiscard() {
+	stateStackLen := len(context.stateStack)
+	if stateStackLen == 0 {
+		return
+	}
+
+	context.stateStack = context.stateStack[:stateStackLen-1]
+}
+
+func (context *meteringContext) ClearStateStack() {
+	context.stateStack = make([]*meteringContext, 0)
+}
+
+func (context *meteringContext) InitStateFromContractCallInput(input *vmcommon.ContractCallInput) {
+	context.unlockGasIfAsyncCallback(input)
+	context.initialGasProvided = input.GasProvided
+}
+
+// unlockGasIfAsyncCallback unlocks the locked gas if the call type is async callback
+func (context *meteringContext) unlockGasIfAsyncCallback(input *vmcommon.ContractCallInput) {
+	if input.CallType != vmcommon.AsynchronousCallBack {
+		return
+	}
+
+	gasProvided := math.AddUint64(input.GasProvided, input.GasLocked)
+
+	input.GasProvided = gasProvided
+	input.GasLocked = 0
 }
 
 // GasSchedule returns the current gas schedule
@@ -80,6 +143,29 @@ func (context *meteringContext) GasLeft() uint64 {
 	}
 
 	return gasProvided - gasUsed
+}
+
+// GasForwarded returns the amount of gas used by the current contract for the
+// execution of other contracts
+func (context *meteringContext) GasForwarded() uint64 {
+	return context.gasForwarded
+}
+
+// ForwardGas accumulates the gas forwarded by the current contract for the execution of other contracts
+func (context *meteringContext) ForwardGas(gas uint64) {
+	context.gasForwarded = math.AddUint64(context.gasForwarded, gas)
+}
+
+func (context *meteringContext) GasUsedByContract() uint64 {
+	runtime := context.host.Runtime()
+	compilationGas := math.SubUint64(context.initialGasProvided, runtime.GetVMInput().GasProvided)
+	executionGasUsed := runtime.GetPointsUsed()
+
+	gasUsed := uint64(0)
+	gasUsed = math.AddUint64(compilationGas, executionGasUsed)
+	gasUsed = math.SubUint64(gasUsed, context.gasForwarded)
+
+	return gasUsed
 }
 
 // BoundGasLimit returns the gas left if it is less than the given limit, or the given value otherwise.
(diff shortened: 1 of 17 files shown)


arwen/contexts/metering_test.go
```diff
@@ -243,8 +243,114 @@ func TestMeteringContext_AsyncCallGasLocking(t *testing.T) {
...
+	require.Equal(t, uint64(1000), metering.GasLeft())
+
+	metering.UseGas(400)
+	require.Equal(t, uint64(600), metering.GasLeft())
+
+	gasUsedByContract := metering.GasUsedByContract()
+	require.Equal(t, uint64(1400), gasUsedByContract)
+}
+
+func TestMeteringContext_GasUsed_StackOneLevel(t *testing.T) {
```
sasurobert, line 284: can you add a test with a few levels of stack ?

arwen/contexts/metering.go
```diff
@@ -82,6 +145,29 @@ func (context *meteringContext) GasLeft() uint64 {
...
+}
+
+// ForwardGas accumulates the gas forwarded by the current contract for the execution of other contracts
+func (context *meteringContext) ForwardGas(gas uint64) {
+	context.gasForwarded = math.AddUint64(context.gasForwarded, gas)
+}
+
+func (context *meteringContext) GasUsedByContract() uint64 {
+	runtime := context.host.Runtime()
+	compilationGas := math.SubUint64(context.initialGasProvided, runtime.GetVMInput().GasProvided)
```
sasurobert, line 161: having code like this makes it easy to use it wrongly. The same issue when the current execution / compilation cost is effectively lost. version 0.4, from 0.5 onwards this was corrected.

sasurobert: there is no instance stacking in the mainnet currently.

arwen/host/asyncCall.go
```diff
@@ -166,7 +168,9 @@ func (host *vmHost) sendCallbackToCurrentCaller() error {
 		return err
 	}
 
-	metering.UseGas(metering.GasLeft())
+	gasLeft := metering.GasLeft()
+	metering.ForwardGas(gasLeft)
```
sasurobert, line 172: forward gas could simply use gas inside.

arwen/host/execution_vmoutputs_test.go
```diff
@@ -345,7 +348,7 @@ func expectedVMOutputSameCtxRecursiveMutualMethods(_ []byte, recursiveCalls int)
 	)
 	account.Balance = big.NewInt(1000)
 	account.BalanceDelta = big.NewInt(0).Sub(big.NewInt(1), big.NewInt(1))
-	account.GasUsed = 25412
+	account.GasUsed = 67668
```
sasurobert, line 351: why were these values changed ?

arwen/host/execution_vmoutputs_test.go
```diff
@@ -820,6 +833,7 @@ func expectedVMOutputDestCtxByCallerSimpleTransfer(value int64) *vmcommon.VMOutp
 	)
 	childAccount.Balance = big.NewInt(1000)
 	childAccount.BalanceDelta = big.NewInt(-value)
+	childAccount.GasUsed = 666
```
sasurobert, line 836: this is an interesting number

arwen/contexts/metering.go
```diff
@@ -26,14 +29,74 @@ func NewMeteringContext(
...
+
 	return context, nil
 }
 
+func (context *meteringContext) InitState() {
+	context.initialGasProvided = 0
+	context.gasForwarded = 0
+}
+
+func (context *meteringContext) PushState() {
```
sasurobert, line 48: missing comments on public functions

arwen/contexts/metering.go
sasurobert: compilation / AoT prepare gas could be save in a separate field. we have to be carefull about backward compatibility

sasurobert (changes requested):

mock/context/meteringContextMock.go
```diff
@@ -45,6 +66,24 @@ func (m *MeteringContextMock) GasLeft() uint64 {
...
+// ForwardGas mocked method
+func (m *MeteringContextMock) ForwardGas(_ uint64) {
+}
+
+// InitStateFromContractCallInput mocked method
+func (m *MeteringContextMock) InitStateFromContractCallInput(input *vmcommon.ContractCallInput) {
+}
+
+// MeteringContextMock mocked method
+func (context *MeteringContextMock) GasUsedByContract() uint64 {
```
BeniaminDrasovean, line 83: m instead of context

BeniaminDrasovean (changes requested):

arwen/host/execution_test.go
```diff
@@ -1285,6 +1280,123 @@ func TestExecution_CreateNewContract_Fail(t *testing.T) {
...
 }
 
+func TestExecution_Mocked_Wasmer_Instances(t *testing.T) {
+	host, _, ibm := defaultTestArwenForCallWithInstanceMocks(t)
+
+	parentInstance := ibm.CreateAndStoreInstanceMock(parentAddress)
+	parentInstance.Exports["callChild"] = mockMethod(func() {
+		host.Output().Finish([]byte("parent returns this"))
+		host.Metering().UseGas(500)
+		host.Storage().SetStorage([]byte("parent"), []byte("parent storage"))
```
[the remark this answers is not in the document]
camilbancioiu: Fixed.

arwen/host/execution_test.go
```diff
@@ -1285,6 +1280,123 @@ func TestExecution_CreateNewContract_Fail(t *testing.T) {
...
+		childInput.GasProvided = 1000
+		_, _, err := host.ExecuteOnDestContext(childInput)
+		require.Nil(t, err)
+	})
+
+	childInstance := ibm.CreateAndStoreInstanceMock(childAddress)
+	childInstance.Exports["doSomething"] = mockMethod(func() {
+		host.Output().Finish([]byte("child returns this"))
+		host.Metering().UseGas(100)
+		host.Storage().SetStorage([]byte("child"), []byte("child storage"))
```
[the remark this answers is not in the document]
camilbancioiu: Fixed.

mock/context/meteringContextMock.go
camilbancioiu: Fixed.

arwen/host/execution_vmoutputs_test.go
camilbancioiu: In this specific case, it's because the same contract was calling itself with `executeOnSameContext()`, so it simply used more gas. The test currently instructs the contract to call itself 5 times, which uses 67668 gas units, but here are more values:

2 recursive calls: 25304 = 2 compilations + 2 executions
3 recursive calls: 37998 = 3 compilations + 3 executions
4 recursive calls: 50730 = 4 compilations + 4 executions
5 recursive calls: 67668 = 5 compilations + 5 executions
6 recursive calls: 84644 = 6 compilations + 6 executions

The previous value was computed differently.

arwen/host/execution_vmoutputs_test.go
camilbancioiu: If you simultaneously roll 3 dice, you'll statistically get the combination "6 6 6" once every 216 throws :P

arwen/contexts/metering.go
camilbancioiu: Added.

arwen/host/asyncCall.go
camilbancioiu: Under the current design, `metering.ForwardGas()` only marks some gas as "forwarded", but the gas itself was already consumed on the Wasmer instance.

In this specific case, for `asyncCall()`, gas forwarding happens when the gas is consumed, but this only happens for cross-shard calls. For in-shard calls, gas is used separately by `host.executeSmartContractCall()`, and later is marked as "forwarded" in `host.finishExecuteOnSame/DestContext()". But the gas marked as "forwarded" is not the entire gas - it's the actual used gas, after unspent gas is restored to the caller.

arwen/contexts/metering.go
camilbancioiu: Will make a separate field.

sasurobert (approved):

BeniaminDrasovean (approved):

PR #223, opened 2020-12-16
nonPayableFuncCheck hook

andrei-marinica:
New hook that checks that no payment was sent.
Checks both EGLD and ESDT.
Updated mandos tests to use it.


arwen/elrondapi/elrondei.go
```diff
@@ -22,6 +22,7 @@ package elrondapi
 // extern int32_t storageLoad(void *context, int32_t keyOffset, int32_t keyLength , int32_t dataOffset);
 // extern int32_t storageLoadFromAddress(void *context, int32_t addressOffset, int32_t keyOffset, int32_t keyLength , int32_t dataOffset);
 // extern void getCaller(void *context, int32_t resultOffset);
+// extern void nonPayableFuncCheck(void *context);
```
andrei-marinica, line 25: If you have a better name in mind, it's not too late to change

BeniaminDrasovean: Maybe checkNonPayableFunc?

BeniaminDrasovean (approved):

arwen/elrondapi/elrondei.go
andrei-marinica: Yeah ... thought about that, but I don't want it to seem like we are checking whether or not the function is non payable. We are checking that no payment came.

sasurobert: I would say checkFunctionNonPayable

arwen/elrondapi/elrondei.go
```diff
@@ -1002,6 +1008,27 @@ func getCaller(context unsafe.Pointer, resultOffset int32) {
...
+//export nonPayableFuncCheck
+func nonPayableFuncCheck(context unsafe.Pointer) {
+	runtime := arwen.GetRuntimeContext(context)
+	metering := arwen.GetMeteringContext(context)
+
+	gasToUse := metering.GasSchedule().ElrondAPICost.GetCallValue
+	metering.UseGas(gasToUse)
+
+	vmInput := runtime.GetVMInput()
+	if vmInput.CallValue.Sign() > 0 {
```
sasurobert, line 1020: vmInput.CallValue.Cmp(zero) > 0 as with ESDTValue -> but for esdt check for nil first.

arwen/elrondapi/elrondei.go
BeniaminDrasovean: Right, makes sense!

arwen/elrondapi/elrondei.go
```diff
@@ -1002,6 +1008,27 @@ func getCaller(context unsafe.Pointer, resultOffset int32) {
...
+		runtime := arwen.GetRuntimeContext(context)
+		arwen.WithFault(arwen.ErrNonPayableFunctionEgld, context, runtime.ElrondAPIErrorShouldFailExecution())
+		return
+	}
+	if vmInput.ESDTValue.Sign() > 0 {
```
sasurobert, line 1025: if vmInput.EsdtValue != nil && vmInput.ESDTValue.Cmp(zero) > 0

sasurobert (changes requested):

sasurobert (approved):

BeniaminDrasovean (approved):

PR #225, opened 2020-12-21
Instance handler

BeniaminDrasovean:
Use wasm.InstanceHandler instead of *wasm.Instance.

### DIFF
--- a/arwen/contexts/runtime.go
+++ b/arwen/contexts/runtime.go
@@ -19,7 +19,7 @@ var _ arwen.RuntimeContext = (*runtimeContext)(nil)
 
 type runtimeContext struct {
 	host         arwen.VMHost
-	instance     *wasmer.Instance
+	instance     wasmer.InstanceHandler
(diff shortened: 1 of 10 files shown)


arwen/contexts/runtime.go
```diff
@@ -257,7 +257,7 @@ func (context *runtimeContext) ResetWarmInstance() {
 		return
 	}
 
-	arwen.RemoveHostContext(*context.instance.Data)
+	arwen.RemoveHostContext(*context.instance.GetData())
```
sasurobert, line 260: is it safe to get the pointer and remove this like this ? make sure GetData returns what it needs.

sasurobert (approved):

ccorcoveanu (approved):

arwen/contexts/runtime.go
BeniaminDrasovean: Tested, and it seems all right.

PR #231, opened 2021-01-13
output-fix-esdt-revert

sasurobert:
gas usage fix for esdt and revert esdt transfer in case of errors

### DIFF
--- a/arwen/host/execution.go
+++ b/arwen/host/execution.go
@@ -570,7 +570,12 @@ func (host *vmHost) execute(input *vmcommon.ContractCallInput) error {
 			runtime.InitStateFromContractCallInput(newVMInput)
 			metering.InitStateFromContractCallInput(&newVMInput.VMInput)
 			storage.SetAddress(runtime.GetSCAddress())
-			return host.executeSmartContractCall(newVMInput, metering, runtime, output, false)
+			err = host.executeSmartContractCall(newVMInput, metering, runtime, output, false)
+			if err != nil {
+				host.revertESDTTransfer(input)
+			}
+
+			return err
 		}
 
 		return nil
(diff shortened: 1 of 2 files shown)


arwen/host/execution.go
```diff
@@ -626,6 +667,12 @@ func (host *vmHost) callBuiltinFunction(input *vmcommon.ContractCallInput) (*vmc
 		return nil, err
 	}
 
+	if newVMInput != nil {
```
SebastianMarian, line 670: Coud be err and newVMInput both nil ?

sasurobert: yes. it is possible.

sasurobert: yes. it is

SebastianMarian (approved):

BeniaminDrasovean (approved):

arwen/host/execution.go
```diff
@@ -596,6 +601,42 @@ func (host *vmHost) callSCMethodIndirect() error {
...
+			Arguments:      input.Arguments[:2],
+			CallValue:      big.NewInt(0),
+			CallType:       vmcommon.DirectCall,
+			GasPrice:       input.GasPrice,
+			GasProvided:    input.GasProvided,
+			GasLocked:      0,
+			OriginalTxHash: input.OriginalTxHash,
+			CurrentTxHash:  input.CurrentTxHash,
+			ESDTValue:      big.NewInt(0),
+			ESDTTokenName:  nil,
```
dorin-iancu, line 624: If this is meant to revert an ESDT transfer, shouldn't this have the original call's ESDTValue and ESDTToken name?

dorin-iancu (approved):

arwen/host/execution.go
sasurobert: it does not matter. built in function does not get it from esdtvalue / esdttokenname but from function and arguments.

PR #236, opened 2021-01-16
check gas at the end

sasurobert:


arwen/host/execution.go
```diff
@@ -686,6 +698,10 @@ func (host *vmHost) callInitFunction() error {
 		err = host.handleBreakpointIfAny(err)
 	}
 
+	if err == nil {
```
iulianpascalau, line 701: return directly?

iulianpascalau (approved):

arwen/host/execution.go
```diff
@@ -710,6 +726,9 @@ func (host *vmHost) callSCMethod() error {
 	if err != nil {
 		err = host.handleBreakpointIfAny(err)
 	}
+	if err == nil {
```
SebastianMarian, line 729: I think L725 - L734 could be refactored like this to be more clear:

	_, err = function()
	if err != nil {
		err = host.handleBreakpointIfAny(err)
	        if err != nil {
		       return err
	        }
	}

	err = host.checkFinalGasAfterExit()
	if err != nil {
		return err
	}

arwen/host/execution.go
SebastianMarian: It's not work to return directly, as if err != nil in this point it should be return this err and if you return directly checkFinalGasAfterExit it could be return nil

SebastianMarian: I also propose this refactor for L696-705 :

	_, err := init()
	if err != nil {
		err = host.handleBreakpointIfAny(err)
                if err != nil {
                       return err
                }
	}

        return host.checkFinalGasAfterExit()

arwen/host/execution.go
```diff
@@ -686,6 +698,10 @@ func (host *vmHost) callInitFunction() error {
 		err = host.handleBreakpointIfAny(err)
```
SebastianMarian, line 698: If init() returns err this call could overwrite the init error with nil, it is ok?

arwen/host/execution.go
```diff
@@ -710,6 +726,9 @@ func (host *vmHost) callSCMethod() error {
 	if err != nil {
 		err = host.handleBreakpointIfAny(err)
```
SebastianMarian, line 727: If function() returns err this call could overwrite the init error with nil, it is ok?

sasurobert: I would leave like this. It is the correct way to do it.

arwen/host/execution.go
sasurobert: I would leave like this. It is the correct way to do it.

arwen/host/execution.go
sasurobert: I would leave like this. It is the correct way to do it.

arwen/host/execution.go
sasurobert: Yes it is correct

arwen/host/execution.go
sasurobert: The code is correct

SebastianMarian (approved):

PR #243, opened 2021-01-21
test memory leakage

sasurobert:
resolved memory leakage when using Cache from a wasmer instance

### DIFF
--- a/wasmer/instance.go
+++ b/wasmer/instance.go
@@ -223,11 +223,8 @@ func NewInstanceFromCompiledCodeWithOptions(
 		instance.InstanceCtx = IntoInstanceContextDirect(c_instance_context)
 	}
 
-	hackMap := make(map[unsafe.Pointer]interface{})
-	hackMap[pointerToDestroy] = struct{}{}
-
-	delete(hackMap, pointerToDestroy)
 	C.free(pointerToDestroy)
+	pointerToDestroy = nil
 
 	return instance, err
 }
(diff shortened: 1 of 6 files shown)


andrei-marinica (approved):

wasmer/instance.go
```diff
@@ -220,6 +222,13 @@ func NewInstanceFromCompiledCodeWithOptions(
 		c_instance_context := cWasmerInstanceContextGet(c_instance)
 		instance.InstanceCtx = IntoInstanceContextDirect(c_instance_context)
 	}
+
+	hackMap := make(map[unsafe.Pointer]interface{})
```
AdoAdoAdo, line 226: why is the hackMap needed?
aren't just lines 206 and 230 relevant?

wasmer/instance.go
```diff
@@ -280,8 +289,12 @@ func (instance *Instance) Cache() ([]byte, error) {
 	}
 
 	goBytes := C.GoBytes(unsafe.Pointer(cacheBytes), C.int(cacheLen))
-	cacheBytes = nil
 
+	hackMap := make(map[unsafe.Pointer]interface{})
```
iulianpascalau, line 293: here too? Why do we need the hack map?

sasurobert: testing now - as without it is fails in some scenarios.

wasmer/instance.go
sasurobert: testing now - as without it is fails in some scenarios.

dorin-iancu (dismissed):

wasmer/instance.go
```diff
@@ -1,5 +1,6 @@
 package wasmer
 
+// #include <stdlib.h>
```
iulianpascalau, line 3: please remove comment

iulianpascalau (commented):
Please add description

wasmer/instance.go
sasurobert: this is not a comment. this is a c include - it is a must for C.free

iulianpascalau: Ok

AdoAdoAdo (approved):

andrei-marinica (approved):

PR #247, opened 2021-02-02
Verify secp256k1 fix

andrei-marinica:
Signature length was hardcoded instead of given as argument.
Now taken from the signature's second byte.


arwen/cryptoapi/cryptoei.go
```diff
@@ -266,7 +267,7 @@ func verifySecp256k1(
 		return 1
 	}
 
-	sig, err := runtime.MemLoad(sigOffset, secp256k1SignatureLength)
+	sig, err := runtime.MemLoad(sigOffset, sigLength)
```
camilbancioiu, line 270: According to the code of `crypto.VerifySecp256k1()`, `sigLength` will be read from `sigOffset + 1` automatically, as a `byte` (the maximum length is 255). We could replace the new `sigLength` parameter of `verifySecp256k1()` with simply reading the byte at `sigOffset + 1`.

camilbancioiu (approved):

ccorcoveanu (approved):

PR #263, opened 2021-03-05
Test for ReturnData API

camilbancioiu:
This PR adds a test for the EEI methods that retrieve the return data. Also used to retrieve data returned by a child contract called using `executeOnDestContext()` and related sync calling methods.

TODO also test using `executeOnSameContext()`
TODO write test that retrieves multiple values from the child execution.


dorin-iancu (approved):
:mage:

claudiu725 (approved):

PR #266, opened 2021-03-09
Mandos real gas

andrei-marinica:
Mandos tests can specify the gas schedule.

Ping-pong-egld contract tests added, which contain txs that interrupt the operation when running low on gas.

Not all tests converted to the new model, this is a lower prio task for later this week.


mandos-go/json/model/gasSchedule.go
```diff
@@ -0,0 +1,19 @@
...
+// GasSchedule encodes the gas model to be used in mandos tests
+type GasSchedule int
+
+const (
+	// GasScheduleDefault indicates that the mandos scenario should use whatever the default gas model is.
+	// Should be the latest version of the mainnet gas schedule.
+	GasScheduleDefault GasSchedule = iota
+
+	// GasScheduleDummy is a dummy model, with all costs set to 1.
+	GasScheduleDummy GasSchedule = iota
```
sasurobert, line 12: why do you have iota 2 times ?

sasurobert (changes requested):

mandos-go/json/model/gasSchedule.go
andrei-marinica: bad copy paste, fixing now

sasurobert (approved):

camilbancioiu (approved):

PR #278, opened 2021-04-07
Copy txhashes in context

sasurobert:


andrei-marinica (approved):

arwen/host/execution.go
```diff
@@ -180,6 +180,20 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
 	return
 }
 
+func copyTxHashesFromContext(copyEnabled bool, runtime arwen.RuntimeContext, input *vmcommon.ContractCallInput) {
```
camilbancioiu, line 183: This could be a method of `runtimeContext`, where the `host` is already available, so `host.IsESDTFunctionsEnabled()` can be called straightforwardly.

Something like `runtime.CopyTxHashesToInput(input)`, everything contained within.

arwen/host/execution.go
```diff
@@ -180,6 +180,20 @@ func (host *vmHost) doRunSmartContractCall(input *vmcommon.ContractCallInput) (v
...
+		return
+	}
+	currentVMInput := runtime.GetVMInput()
+	if len(currentVMInput.OriginalTxHash) > 0 {
+		input.OriginalTxHash = currentVMInput.OriginalTxHash
+	}
+
+	if len(currentVMInput.CurrentTxHash) > 0 {
+		input.CurrentTxHash = currentVMInput.CurrentTxHash
+	}
```
camilbancioiu, line 194: Should also copy `input.PrevTxHash`.

dorin-iancu (approved):

arwen/host/execution.go
sasurobert: copied.

arwen/host/execution.go
sasurobert: it is so particular that I would not make a public function and interface for it.

camilbancioiu (approved):

arwen/host/execution.go
camilbancioiu: Will eventually move it on the `host` then, as a private method.

PR #279, opened 2021-04-08
Fix gas consumed on transfer and exec

sasurobert:


bogdan-rosianu (approved):

arwen/host/execution.go
```diff
@@ -740,6 +740,11 @@ func (host *vmHost) ExecuteESDTTransfer(destination []byte, sender []byte, token
 	}
 
 	gasConsumed, _ := math.SubUint64(esdtTransferInput.GasProvided, vmOutput.GasRemaining)
+	for _, outAcc := range vmOutput.OutputAccounts {
```
iulianpascalau, line 743: could have extracted in a new function

iulianpascalau (approved):
I do not think those are quite backwards compatible with the current testnet :D

PR #300, opened 2021-05-12
fix

dorin-iancu:
Fixing and issue in mandos-go where an NFT with unspecified hash would throw an error like "want nil hash, have [0 0 0 ... 0 0]". This makes it so an empty hash in the "real" NFT doesn't throw this error.


andrei-marinica (approved):

andrei-marinica:
Feature already implemented in https://github.com/ElrondNetwork/arwen-wasm-vm/pull/301

PR #307, opened 2021-05-24
arwenmandos - gas schedules embedded in binary

andrei-marinica:
Additional fix: some test files were getting loaded in the executable too, causing it to crash.


arwenmandos/gasSchedules/scripts/includetoml.go
```diff
@@ -0,0 +1,27 @@
+package main
```
camilbancioiu, line 1: Consider moving this file to`cmd/includetoml/main.go`. All executable program should be in the same place. The file produced by `main()` uses a relative path anyway.

camilbancioiu (commented):
Good idea.

arwenmandos/gasSchedules/scripts/includetoml.go
andrei-marinica: It is only a macro, only used in the `go:generate` in its parent.

andrei-marinica: `cmd` contains the public binaries

camilbancioiu (approved):

matei-p (approved):

PR #317, opened 2021-06-08
IsSmartContract - Make it work in #init function

matei-p:


andrei-marinica (dismissed):

arwen/hosttest/execution_test.go
```diff
@@ -2267,6 +2268,45 @@ func TestExecution_CreateNewContract_Fail(t *testing.T) {
 		})
 }
 
+func TestExecution_CreateNewContract_IsSmartContract(t *testing.T) {
+
+	childCode := test.GetTestSCCode("deployer-child", "../../")
+
+	ownerNonce := uint64(23)
+	parentAddress := testcommon.MakeTestSCAddress("newAddr_" + fmt.Sprint(24))
+	childAddress := testcommon.MakeTestSCAddress("newAddr_" + fmt.Sprint(25))
```
andrei-marinica, line 2277: I usually write something like this, how it is here is also fine
```suggestion
	childAddress := testcommon.MakeTestSCAddress(fmt.Sprintf("newAddr_%d", 25))
```

andrei-marinica (dismissed):

test/contracts/deployer-child/elrond.json
```diff
@@ -0,0 +1,6 @@
+{
+    "language": "clang",
+    "source_files": [
+        "/home/bogdan/Elrond/arwen-wasm-vm/test/contracts/deployer-child/deployer-child.c"
```
sasurobert, line 4: source file is wrong, like path should be not under /home/bogdan

arwen/elrondapi/elrondei.go
```diff
@@ -520,6 +521,12 @@ func v1_3_isSmartContract(context unsafe.Pointer, addressOffset int32) int32 {
 	}
 
 	isSmartContract := blockchain.IsSmartContract(address)
+
+	if !isSmartContract {
```
sasurobert, line 525: this code is not needed. The IsSmartContract feature is not working properly on the WorldMock - in case of Elrond - blockchain hook it only verifies the type of the address. The worldMock makes something else.
The current code would return true even for a normal address which had a storage update.

sasurobert (changes requested):

test/contracts/deployer-child/elrond.json
matei-p: hmmm is possible that the build tool put it here

test/contracts/deployer-parent/deployer-parent.c
```diff
@@ -0,0 +1,43 @@
...
+			deploymentValue,
+			childCode,
+			childMetadata,
+			codeSize,
+			childGeneratedAddress,
+			lastArg + 1,
+			(byte*)args.lengthsAsI32,
+			args.serialized);
+
+	finishResult(result);
```
camilbancioiu, line 42: Add a verification here as well, check if the deployed child is a smart contract.

camilbancioiu (changes requested):

camilbancioiu (dismissed):

camilbancioiu (approved):

andrei-marinica (approved):

sasurobert (approved):

PR #318, opened 2021-06-08
error reporting fix

matei-p:


arwen/contexts/runtime.go
```diff
@@ -518,7 +518,12 @@ func (context *runtimeContext) FailExecution(err error) {
 
 	var message string
 	if err != nil {
-		message = err.Error()
+		wrappedError, ok := (err.(arwen.WrappableError))
```
iulianpascalau, line 521: can be simplified as:
```
wrappedError, ok := err.(arwen.WrappableError)
```

matei-p: err is not always a WrappableError if this is the assumption

sasurobert (approved):

arwen/contexts/runtime.go
sasurobert: the paranthesis are duplicated.

iulianpascalau: no, please remove the redundant parenthesis around the type assertion. It raises linter issues.

iulianpascalau (approved):

arwen/contexts/runtime.go
```diff
@@ -531,7 +531,7 @@ func (context *runtimeContext) FailExecution(err error) {
 	context.host.Output().SetReturnMessage(message)
 	context.SetRuntimeBreakpointValue(arwen.BreakpointExecutionFailed)
 
-	logRuntime.Trace("execution failed", "message", message)
+	logRuntime.Trace("execution failed", "message", err.Error())
```
iulianpascalau, line 534: this can panic

iulianpascalau (changes requested):

arwen/contexts/runtime.go
sasurobert: better add an if before trace. this call is not really called with err = nil. but it is better to be safe

iulianpascalau (approved):

sasurobert (approved):

PR #324, opened 2021-06-14
forwarder fuzzer

andrei-marinica:


fuzz/forwarder/fuzzForwarder_test.go
```diff
@@ -0,0 +1,84 @@
...
+		if nonce > 0 {
+			for toIndex == fromIndex {
+				toIndex = r.Intn(pfe.data.numForwarders) + 1
+			}
+		}
+		err = pfe.programCall(callType, fromIndex, toIndex, tokenName, nonce, "10")
+		require.Nil(t, err)
+	}
+
+	err = pfe.executeCallCheckLogs(1)
```
catalinnnn, line 82: Shouldn't this be inside the for loop?

fuzz/forwarder/fuzzForwarderCall.go
```diff
@@ -0,0 +1,139 @@
...
+		"expect": {
+			"out": [],
+			"status": "",
+			"logs": [%s
+			]
+		}
+	`,
+		expectedLogs,
+	)
+	_, err = pfe.executeCall(toIndex, expectLogs)
```
catalinnnn, line 39: I think you should also check if the call did what it was supposed to do. It's good that we can know when sync/async/transfer execute call breaks arwen but it would be nice if we would also detect cases where async call (for example) did not transfer the value, or trasfer execute only transferred the value and exited silently without executing the function

fuzz/forwarder/fuzzForwarderCall.go
```diff
@@ -0,0 +1,139 @@
...
+
+func (pfe *fuzzExecutor) executeCallPrintLogs(toIndex int) error {
+	el, _ := pfe.getAllExpectedLogs(toIndex)
+	fmt.Println(el)
+	output, err := pfe.executeCall(toIndex, "")
+	if err != nil {
+		return err
+	}
+	for _, log := range output.Logs {
+		fmt.Println(pfe.arwenTestExecutor.LogToString(log))
```
catalinnnn, line 18: Maybe we can enforce the tests to also contain the logs that we exected

PR #325, opened 2021-06-15
Hotfix nil worldmock

camilbancioiu:
This PR fixes the `MockWorld` snapshot implementation. Also ensures that the error returned by `handleAsyncCall()` appears in the `RuntimeContext`.


arwen/host/breakpoints.go
```diff
@@ -10,13 +10,13 @@ func (host *vmHost) handleBreakpointIfAny(executionErr error) error {
...
+
 	breakpointValue := runtime.GetRuntimeBreakpointValue()
 	if breakpointValue != arwen.BreakpointNone {
-		if breakpointValue != arwen.BreakpointAsyncCall {
-			runtime.AddError(executionErr, runtime.Function())
-			return host.handleBreakpoint(breakpointValue)
-		}
-		return host.handleBreakpoint(breakpointValue)
+		err := host.handleBreakpoint(breakpointValue)
+		runtime.AddError(err)
```
matei-p, line 18: nice

matei-p (dismissed):

mock/world/mockESDTAccount.go
```diff
@@ -149,8 +149,8 @@ func (a *Account) SetTokenData(tokenKey []byte, tokenData *esdt.ESDigitalToken)
 	if err != nil {
 		return err
 	}
-
-	return a.DataTrieTracker().SaveKeyValue(tokenKey, marshaledData)
+	a.Storage[string(tokenKey)] = marshaledData
```
sasurobert, line 152: do we still need mockESDTAccount ? if yes, I think it was a little bit safer to use the SaveKeyValue as there is the Journal increased for ESDT transfers, even in mocks.

arwen/host/breakpoints.go
```diff
@@ -10,13 +10,13 @@ func (host *vmHost) handleBreakpointIfAny(executionErr error) error {
 	}
 
 	runtime := host.Runtime()
+	runtime.AddError(executionErr)
```
sasurobert, line 13: I think the code before was safer. Like we do not need to add error in case of asyncCall breakpoint. and execution error is not important than.

mock/world/mockESDTAccount.go
camilbancioiu: `SaveKeyValue` ended up being called before the test was even started (during test setup). This led to snapshots being created before the test, causing Mandos rollbacks to have strange behavior and tests were failing for no apparent reason.

Avoiding `SaveKeyValue` is safer here, because the functions in `mockESDTAccount.go` are only called during test setup, when they must leave the journal unaffected.

arwen/host/breakpoints.go
camilbancioiu: Agree about `executionErr`, will try to find a way to make certain tests pass without this line.

But are we sure we do not want the error returned by `handleAsyncCallBreakpoint()` below? Previously it was ignored, but now it is added.

sasurobert: The second part of the changes are good. but there is no need to AddError(executionErr) in case of breakpoiintAsyncCall

camilbancioiu: Removed.

matei-p (approved):

sasurobert (approved):

PR #326, opened 2021-06-15
Elrond ethereum bridge fuzz

dorin-iancu:
Fuzzer for Elrond-Ethereum bridge


fuzz/elrond-ethereum-bridge/fuzzElrondEthereumBridge_test.go
```diff
@@ -0,0 +1,390 @@
...
+
+	fe := newExecutorWithPaths()
+	defer fe.saveGeneratedScenario()
+
+	err := fe.initData()
+	if err != nil {
+		t.Error(err)
+	}
+
+	// TODO: Uncomment once aggregator is integrated
```
catalinnnn, line 75: I think that aggregator was integrated? Is this TODO up to date?

fuzz/elrond-ethereum-bridge/fuzzElrondEthereumBridge_test.go
```diff
@@ -0,0 +1,390 @@
...
+	r.Seed(seed)
+
+	fe.randSource = *r
+
+	re := fuzzutil.NewRandomEventProvider(r)
+	for stepIndex := 1; stepIndex <= *iterationsFlag; stepIndex++ {
+		re.Reset()
+
+		switch {
+		case re.WithProbability(0.2):
```
catalinnnn, line 136: Please consider using Roulette implementation. It's cleaner and more extensible. Like the case when you would want to add a bunch of other events, with the current re.WithProbability, you would have to change all current probabilities and make sure the new sum does not surpass 1. With Roulette this would not be necessary

PR #330, opened 2021-06-17
upgrade contract from contract fix

andrei-marinica:


arwen/contexts/metering.go
```diff
@@ -213,7 +213,7 @@ func (context *meteringContext) checkGas(vmOutput *vmcommon.VMOutput) error {
 	gasProvided := context.GetGasProvided()
 
 	if totalGas != gasProvided {
-		logOutput.Error("gas usage mismatch", "total gas", totalGas, "gas provided", gasProvided)
+		logOutput.Trace("gas usage mismatch", "total gas", totalGas, "gas provided", gasProvided)
 		return arwen.ErrInputAndOutputGasDoesNotMatch
```
sasurobert, line 217: This should have stayed as error in order to catch other cases.

dorin-iancu (dismissed):

catalinnnn (approved):

arwen/elrondapi/elrondei.go
```diff
@@ -1380,15 +1381,12 @@ func v1_3_upgradeContract(
 		return
 	}
 
-	gasSchedule := metering.GasSchedule()
-	gasToUse = gasSchedule.ElrondAPICost.AsyncCallStep
```
iulianpascalau, line 1384: Are we sure we do not want to consume the async call gas?

arwen/elrondapi/elrondei.go
```diff
@@ -1380,15 +1381,12 @@ func v1_3_upgradeContract(
...
-	metering.UseGas(gasToUse)
```
sasurobert, line 1385: API call cost should have stayed

arwen/elrondapi/elrondei.go
camilbancioiu: We are consuming it, but properly. See `runtime.ExecuteAsyncCall()` called below.

arwen/elrondapi/elrondei.go
```diff
@@ -1402,23 +1400,20 @@ func v1_3_upgradeContract(
...
-		Destination: calledSCAddress,
-		Data:        []byte(finalData),
-		GasLimit:    uint64(gasLimit),
-		ValueBytes:  value,
-	})
-
-	// Instruct Wasmer to interrupt the execution of the caller SC.
-	runtime.SetRuntimeBreakpointValue(arwen.BreakpointAsyncCall)
+	runtime.ExecuteAsyncCall(
+		calledSCAddress,
```
sasurobert, line 1413: Reading it again and I think this should have been handled with ExecuteOnDestContext. Not with asyncCall

iulianpascalau (approved):

arwen/elrondapi/elrondei.go
andrei-marinica: the contract could in principle be in another shard, if "change owner" was called

arwen/elrondapi/elrondei.go
camilbancioiu: See `runtime.ExecuteAsyncCall()` below, it consumes the `AsyncCallStep` cost.

arwen/elrondapi/elrondei.go
camilbancioiu: `ExecuteOnDestContext` is safe as long as we don't migrate accounts between shards.

sasurobert (approved):

arwen/elrondapi/elrondei.go
andrei-marinica: I mean, the contract could have its ownership transferred to an account from another shard. The new owner will be the one making the update

sasurobert (changes requested):
Too much revert. Please keep error for Arwen coding errors. For user errors keep trace

sasurobert (approved):

iulianpascalau (approved):

PR #356, opened 2021-07-19
Merge v1.3.24 in man buffers

andrei-diaconescu:


sasurobert (changes requested):
you should merge code from master. not from v1.3.24

PR #357, opened 2021-07-20
Managed Buffers - Refactor, APIs, integrationTest 

andrei-diaconescu:


arwen/elrondapi/manBufOps.go
```diff
@@ -0,0 +1,415 @@
...
+// #include <stdlib.h>
+// typedef unsigned char uint8_t;
+// typedef int int32_t;
+//
+// extern int32_t	v1_4_mBufferNew(void* context);
+// extern int32_t 	v1_4_mBufferNewFromBytes(void* context, int32_t dataOffset, int32_t dataLength);
+// extern int32_t	v1_4_mBufferSetBytes(void* context, int32_t mBufferHandle, int32_t dataOffset, int32_t dataLength);
+// extern int32_t 	v1_4_mBufferGetLength(void* context, int32_t mBufferHandle);
+// extern int32_t	v1_4_mBufferGetBytes(void* context, int32_t mBufferHandle, int32_t resultOffset);
+// extern int32_t	v1_4_mBufferExtendFromSlice(void* context, int32_t mBufferHandle, int32_t dataOffset, int32_t dataLength);
```
sasurobert, line 14: mBufferAppend ?

arwen/elrondapi/manBufOps.go
```diff
@@ -0,0 +1,415 @@
...
+
+	gasToUse := metering.GasSchedule().ManagedBufferAPICost.MBufferSetBytes
+	metering.UseGas(gasToUse)
+
+	data, err := runtime.MemLoad(dataOffset, dataLength)
+	if arwen.WithFault(err, context, runtime.ManagedBufferAPIErrorShouldFailExecution()) {
+		return 1
+	}
+
+	isSuccess := managedType.AppendBytesToThisManagedBuffer(mBufferHandle, data)
```
sasurobert, line 223: The function name could be shorter for the managerType context - like a simple AppendBytes - there are other functions which are too long.

sasurobert: there is no need to have in the function name "ToThisManagedBuffer"

arwen/contexts/managedType.go
```diff
@@ -313,103 +306,103 @@ func (context *managedTypesContext) GetPrivateKeyByteLengthEC(ecHandle int32) in
...
+	mBufferHandle := context.NewManagedBuffer()
+	context.SetBytesForThisManagedBuffer(mBufferHandle, bytes)
+	return mBufferHandle
 }
 
 // SetBytesForThisManagedBuffer sets the bytes given as value for the managed buffer. Returns 0 if success, 1 otherwise
-func (context *managedTypesContext) SetBytesForThisManagedBuffer(manBufHandle int32, bytes []byte) int32 {
-	_, ok := context.manBufValues[manBufHandle]
+func (context *managedTypesContext) SetBytesForThisManagedBuffer(mBufferHandle int32, bytes []byte) bool {
+	_, ok := context.managedTypesValues.mBufferValues[mBufferHandle]
```
sasurobert, line 330: in case of Set functions I think we could work with GetOrCreate - thus creating the managed buffer if it did not exist.
The same with GetArgument to managed buffer

sasurobert (dismissed):

arwen/contexts/managedType.go
andrei-marinica: I am more in favor of only working with explicit buffer creation, since it is safer.
Trying to get or set to a non-existing one should be an error.

andrei-marinica (changes requested):

andrei-marinica (approved):

sasurobert (approved):

PR #386, opened 2021-08-10
pseudo-randomness generator

andrei-diaconescu:


arwen/contexts/managedType.go
```diff
@@ -56,6 +58,19 @@ func NewManagedTypesContext(host arwen.VMHost) (*managedTypesContext, error) {
...
 
+func initRandomizer(host arwen.VMHost) *math.SeedRandReader {
+	randomizer, _ := math.NewSeedRandReader(append(host.Blockchain().CurrentRandomSeed(), host.Runtime().GetCurrentTxHash()...))
+
+	return randomizer
+}
+
+func (context *managedTypesContext) GetRandReader(ctx unsafe.Pointer) *math.SeedRandReader {
+	if context.Reader == nil {
+		context.Reader = initRandomizer(arwen.GetVMHost(ctx))
```
sasurobert, line 69: you do not need to do GetVMHost - you already have access as context.host

sasurobert: So you do not need the unsafe.Pointer as input.

arwen/contexts/managedType.go
```diff
@@ -56,6 +58,19 @@ func NewManagedTypesContext(host arwen.VMHost) (*managedTypesContext, error) {
 	return context, nil
 }
 
+func initRandomizer(host arwen.VMHost) *math.SeedRandReader {
+	randomizer, _ := math.NewSeedRandReader(append(host.Blockchain().CurrentRandomSeed(), host.Runtime().GetCurrentTxHash()...))
+
+	return randomizer
+}
+
+func (context *managedTypesContext) GetRandReader(ctx unsafe.Pointer) *math.SeedRandReader {
```
sasurobert, line 67: add comment to public function

math/randomnessGenerator.go
```diff
@@ -0,0 +1,37 @@
...
+	"crypto/sha256"
+	"encoding/binary"
+	"math/rand"
+)
+
+type SeedRandReader struct {
+	rand *rand.Rand
+}
+
+func NewSeedRandReader(seed []byte) (*SeedRandReader, error) {
```
sasurobert, line 13: add comment to public function

arwen/contexts/managedType.go
```diff
@@ -56,6 +58,19 @@ func NewManagedTypesContext(host arwen.VMHost) (*managedTypesContext, error) {
...
+	if context.Reader == nil {
```
sasurobert, line 68: do check.IfNil - instead of simple nil check here. make reader implement IsInterfaceNil. Make Reader private as well.

arwen/contexts/managedType.go
```diff
@@ -56,6 +58,19 @@ func NewManagedTypesContext(host arwen.VMHost) (*managedTypesContext, error) {
 	return context, nil
 }
 
+func initRandomizer(host arwen.VMHost) *math.SeedRandReader {
+	randomizer, _ := math.NewSeedRandReader(append(host.Blockchain().CurrentRandomSeed(), host.Runtime().GetCurrentTxHash()...))
```
sasurobert, line 62: treat error here. or make NewSeedRandReader to not return error.

math/randomnessGenerator.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(seed) == 0 {
```
sasurobert, line 14: no need for this check.

math/randomnessGenerator.go
```diff
@@ -0,0 +1,37 @@
...
+	rand := rand.New(source)
+
+	return &SeedRandReader{
+		rand: rand,
+	}, nil
+}
+
+// Read will read upto len(p) bytes. It will rotate the existing byte buffer (seed) until it will fill up the provided
+// p buffer
+func (srr *SeedRandReader) Read(p []byte) (n int, err error) {
```
sasurobert, line 31: implement IsInterfaceNil

arwen/contexts/managedType.go
sasurobert: and use and Interface as return here. and keep the context.Reader as interface as well.

math/randomnessGenerator.go
```diff
@@ -0,0 +1,37 @@
...
+		return nil, ErrSeedLengthIsZero
+	}
+
+	seedHash := sha256.Sum256(seed)
+	seedNumber := binary.BigEndian.Uint64(seedHash[:])
+
+	source := rand.NewSource(int64(seedNumber))
+	rand := rand.New(source)
+
+	return &SeedRandReader{
```
sasurobert, line 24: unit tests here and in all new implementation please

sasurobert (changes requested):

arwen/contexts/managedType.go
andrei-diaconescu: it cannot return error here, that's why error is not treated. len(seed) > 0 in this specific use case.

arwen/contexts/managedType.go
```diff
@@ -4,6 +4,7 @@ import (
 	"crypto/elliptic"
 	basicMath "math"
 	"math/big"
+	"unsafe"
```
iulianpascalau, line 7: do not use unsafe

arwen/contexts/managedType.go
```diff
@@ -33,6 +34,7 @@ type managedTypesContext struct {
 	host               arwen.VMHost
 	managedTypesValues managedTypesState
 	managedTypesStack  []managedTypesState
+	Reader             *math.SeedRandReader
```
iulianpascalau, line 37: use interface

math/randomnessGenerator.go
iulianpascalau: is this going to be called in a concurrent manner?

math/randomnessGenerator.go
```diff
@@ -0,0 +1,37 @@
+package math
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"math/rand"
+)
+
+type SeedRandReader struct {
```
iulianpascalau, line 9: make unexported & add tests.
Please do add tests that check that for a predefined string we always get the same n "random" bytes. In this way we have the certainty that we won't add backwards compatibility bugs. (the predefined strings should have some comments stating that those strings should not change)

math/randomnessGenerator.go
```diff
@@ -0,0 +1,37 @@
...
+
+	source := rand.NewSource(int64(seedNumber))
+	rand := rand.New(source)
+
+	return &SeedRandReader{
+		rand: rand,
+	}, nil
+}
+
+// Read will read upto len(p) bytes. It will rotate the existing byte buffer (seed) until it will fill up the provided
```
iulianpascalau, line 29: Are we sure the comment does what it states? Maybe update the comment? (I remember this comment from the first implementation done by me)

math/errors.go
```diff
@@ -12,3 +12,6 @@ var ErrSubtractionUnderflow = errors.New("subtraction underflow")
 
 // ErrMultiplicationOverflow is raised when there is an overflow because of the multiplication of two numbers
 var ErrMultiplicationOverflow = errors.New("multiplication overflow")
+
+// ErrSeedLengthIsZero is raised when length of given seed is 0
+var ErrSeedLengthIsZero = errors.New("seed length cannot be 0")
```
iulianpascalau, line 17: this should be removed when resolving @sasurobert comment

arwen/contexts/managedType.go
```diff
@@ -56,6 +58,19 @@ func NewManagedTypesContext(host arwen.VMHost) (*managedTypesContext, error) {
 	return context, nil
 }
 
+func initRandomizer(host arwen.VMHost) *math.SeedRandReader {
```
iulianpascalau, line 61: + add a unit test. Same here: hardcoded values that will always produce the same outcome. Backwards compatibility and deterministic results are important.

iulianpascalau (changes requested):

math/randomnessGenerator.go
sasurobert: no. never.

arwen/contexts/managedType_test.go
```diff
@@ -29,6 +31,49 @@ func TestNewManagedTypes(t *testing.T) {
...
+	prg := managedTypesContext.GetRandReader()
+	a := make([]byte, 100)
+	prg.Read(a)
+	b := make([]byte, 100)
+	for i := 0; i < 1000; i++ {
+		prg.Read(b)
+		require.NotEqual(t, a, b)
+		fmt.Println(hex.EncodeToString(a))
+		fmt.Println(hex.EncodeToString(b))
+		copy(a, b)
```
iulianpascalau, line 73: why this?

math/interface.go
```diff
@@ -0,0 +1,6 @@
+package math
+
+type RandomnessGenerator interface {
```
iulianpascalau, line 3: Missing comment

math/randomnessGenerator_test.go
```diff
@@ -0,0 +1,44 @@
...
+	require.True(t, firstRandomizer.IsInterfaceNil())
+	firstRandomizer = NewSeedRandReader(firstSeed)
+	require.False(t, firstRandomizer.IsInterfaceNil())
+
+	secondStr := "the usage of the same seed must generate 2 equal randomizers"
+	secondSeed := []byte(secondStr)
+	var secondRandomizer *seedRandReader
+	require.True(t, secondRandomizer.IsInterfaceNil())
+	secondRandomizer = NewSeedRandReader(secondSeed)
+	require.Equal(t, firstRandomizer, secondRandomizer)
```
iulianpascalau, line 22: why this?

math/randomnessGenerator_test.go
```diff
@@ -0,0 +1,44 @@
...
+	var thirdRandomizer *seedRandReader
+	require.True(t, thirdRandomizer.IsInterfaceNil())
+	thirdRandomizer = NewSeedRandReader(thirdSeed)
+	require.NotEqual(t, firstRandomizer, thirdRandomizer)
+	require.NotEqual(t, secondRandomizer, thirdRandomizer)
+	require.False(t, thirdRandomizer.IsInterfaceNil())
+
+	a := make([]byte, 100)
+	firstRandomizer.Read(a)
+	require.NotEqual(t, firstRandomizer, secondRandomizer)
```
iulianpascalau, line 36: can be removed

math/randomnessGenerator_test.go
```diff
@@ -0,0 +1,44 @@
...
+	b := make([]byte, 100)
+	secondRandomizer.Read(b)
+	require.Equal(t, a, b)
+	require.Equal(t, firstRandomizer, secondRandomizer)
```
iulianpascalau, line 40: can be removed

math/randomnessGenerator.go
```diff
@@ -6,32 +6,29 @@ import (
...
-
-// Read will read upto len(p) bytes. It will rotate the existing byte buffer (seed) until it will fill up the provided
-// p buffer
-func (srr *SeedRandReader) Read(p []byte) (n int, err error) {
-	if len(p) == 0 {
-		return 0, ErrSeedLengthIsZero
 	}
+}
 
+// Read generates len(p) random bytes and writes them into p. It always returns len(p) and a nil error.
```
iulianpascalau, line 26: are we sure it always returns nil error? What about if I provide a nil buffer? Please add unit tests for nil buffer & empty buffer. Check what it returns.

iulianpascalau (changes requested):

math/randomnessGenerator.go
iulianpascalau: great :+1:

arwen/contexts/managedType_test.go
iulianpascalau: for the compare, figured out later. All good :+1:

math/randomnessGenerator.go
iulianpascalau: Can add a separate test called "...._BackwardsCompatibilityCheck...."

math/randomnessGenerator_test.go
```diff
@@ -0,0 +1,44 @@
...
+	c := make([]byte, 100)
+	thirdRandomizer.Read(c)
+	require.NotEqual(t, a, c)
+}
```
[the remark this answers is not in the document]
iulianpascalau: Can add a separate test called "...._BackwardsCompatibilityCheck...."

arwen/contexts/managedType.go
```diff
@@ -33,6 +35,7 @@ type managedTypesContext struct {
 	host               arwen.VMHost
 	managedTypesValues managedTypesState
 	managedTypesStack  []managedTypesState
+	srr                math.RandomnessGenerator
```
sasurobert, line 38: change srr to randomnessGenerator

math/randomnessGenerator_test.go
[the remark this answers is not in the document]
sasurobert: There is no need for backward compatibility check - as this is not an active function on mainnet.

sasurobert (changes requested):

sasurobert (approved):

math/randomnessGenerator_test.go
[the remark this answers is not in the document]
iulianpascalau: Will be at some point. I don't want to debug a one liner future bug.

iulianpascalau (approved):

PR #396, opened 2021-08-24
Big float api

andrei-diaconescu:


arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,753 @@
...
+
+	gasToUse := metering.GasSchedule().BigFloatAPICost.BigFloatNew
+	metering.UseGas(gasToUse)
+
+	if exponent >= 0 {
+		_ = arwen.WithFault(arwen.ErrPositiveExponent, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return -1
+	}
+
+	value := float64(intBase) + float64(subIntBase)*math.Pow10(int(exponent))
```
sasurobert, line 214: gas should depend on the length of exponent. Plus limit it to a certain constant

andrei-diaconescu: by default if exponent < -323 , then Pow10(exponent) is equal to 0

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+//
+// extern void		v1_4_bigFloatGetConstPi(void* context, int32_t destinationHandle);
+// extern void		v1_4_bigFloatGetConstE(void* context, int32_t destinationHandle);
+//
+// extern void		v1_4_bigFloatSetBytes(void* context, int32_t destinationHandle, int32_t dataOffset, int32_t dataLength);
+// extern void		v1_4_bigFloatGetBytes(void* context, int32_t destinationHandle, int32_t dataOffset);
+// extern void		v1_4_bigFloatFinish(void* context, int32_t referenceHandle);
+// extern void		v1_4_bigFloatGetArgument(void* context, int32_t id, int32_t destinationHandle);
+//
+// extern int32_t	v1_4_bigFloatStorageStore(void* context, int32_t keyOffset, int32_t keyLength, int32_t sourceHandle);
```
andrei-marinica, line 43: Implement conversions to/from managed buffers, after that specialized storage/argument/result hooks are no longer needed.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+// extern void		v1_4_bigFloatMul(void* context, int32_t destinationHandle, int32_t op1Handle, int32_t op2Handle);
+// extern void		v1_4_bigFloatDiv(void* context, int32_t destinationHandle, int32_t op1Handle, int32_t op2Handle);
+// extern void		v1_4_bigFloatRoundDiv(void* context, int32_t destinationHandle, int32_t op1Handle, int32_t op2Handle);
+// extern void		v1_4_bigFloatMod(void* context, int32_t destinationHandle, int32_t op1Handle, int32_t op2Handle);
+//
+// extern void		v1_4_bigFloatAbs(void* context, int32_t destinationHandle, int32_t opHandle);
+// extern void		v1_4_bigFloatNeg(void* context, int32_t destinationHandle, int32_t opHandle);
+// extern int32_t	v1_4_bigFloatCmp(void* context, int32_t op1Handle, int32_t op2Handle);
+// extern int32_t	v1_4_bigFloatSign(void* context, int32_t opHandle);
+// extern void 		v1_4_bigFloatCopy(void* context, int32_t destinationHandle, int32_t opHandle);
```
andrei-marinica, line 23: Rename to v1_4_bigFloatClone, it is more suggestive.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+
+	gasToUse := metering.GasSchedule().BigFloatAPICost.BigFloatNew
+	metering.UseGas(gasToUse)
+
+	if denominator == 0 {
+		_ = arwen.WithFault(arwen.ErrDivZero, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return -1
+	}
+
+	value := float64(numerator) / float64(denominator)
```
andrei-marinica, line 258: Is this division safe/deterministic?

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+	dest := managedType.GetBigFloatOrCreate(destinationHandle)
+	op1, op2, err := managedType.GetTwoBigFloat(op1Handle, op2Handle)
+	if arwen.WithFault(err, context, runtime.BigFloatAPIErrorShouldFailExecution()) {
+		return
+	}
+	if oneIsInfinity(op1, op2) {
+		_ = arwen.WithFault(arwen.ErrInfinityFloatOperation, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return
+	}
+	if allAreEqualToZero(op1, op2) {
```
andrei-marinica, line 346: Division by zero unhandled! Check op2 separately.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+	basePrecision := base.Prec()
+	expPrecision := exp.Prec()
+	if basePrecision > expPrecision {
+		counter.SetPrec(basePrecision)
+		result.SetPrec(basePrecision)
+	} else {
+		counter.SetPrec(expPrecision)
+		result.SetPrec(expPrecision)
+	}
+	for counter.Cmp(exp) < 0 {
```
andrei-marinica, line 550: This isn't how real number exponentiation works. This is float ^ int.
Rename it to powInt or something and make the second argument int32.
For real number exponentiation you would need some Exp function and combine that with Log2. No need to implement now.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+func v1_4_bigFloatPow(context unsafe.Pointer, destinationHandle, op1Handle, op2Handle int32) {
+	managedType := arwen.GetManagedTypesContext(context)
+	metering := arwen.GetMeteringContext(context)
+	runtime := arwen.GetRuntimeContext(context)
+
+	gasToUse := metering.GasSchedule().BigIntAPICost.BigIntPow
+	metering.UseGas(gasToUse)
+
+	dest := managedType.GetBigFloatOrCreate(destinationHandle)
+	op1, op2, err := managedType.GetTwoBigFloat(op1Handle, op2Handle)
```
andrei-marinica, line 510: op2 can be a regulat small int, I don't think there will ever be enough gas for huge exponentiation. Big int probably overkill, definitely not float, since we are only doing repeated multiplications.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+		_ = arwen.WithFault(arwen.ErrInfinityFloatOperation, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(op)
+	op.Int(bigIntOp)
+	managedType.ConsumeGasForBigIntCopy(bigIntOp)
+	if op.IsInt() {
+		return
+	}
+	if bigIntOp.Sign() < 0 {
```
andrei-marinica, line 581: Isn't here an edge case where the int is actually equal to the float?
floor(-1) = -1
floor(-1.1) = -2

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+		_ = arwen.WithFault(arwen.ErrInfinityFloatOperation, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(op)
+	op.Int(bigIntOp)
+	managedType.ConsumeGasForBigIntCopy(bigIntOp)
+	if op.IsInt() {
+		return
+	}
+	if bigIntOp.Sign() > 0 {
```
andrei-marinica, line 610: Isn't here an edge case where the int is actually equal to the float?
ceil(1) = 1
ceil(1.1) = 2

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
...
+		bigIntOp.Add(bigIntOp, big.NewInt(1))
+	}
+}
+
+//export v1_4_bigFloatSetInt64
+func v1_4_bigFloatSetInt64(context unsafe.Pointer, destinationHandle int32, value int64) {
+	managedType := arwen.GetManagedTypesContext(context)
+	metering := arwen.GetMeteringContext(context)
+
+	gasToUse := metering.GasSchedule().BigFloatAPICost.BigFloatSub
```
andrei-marinica, line 620: gasToUse := metering.GasSchedule().BigFloatAPICost.BigFloatSetInt

andrei-marinica (changes requested):

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,893 @@
+package elrondapi
```
alyn509, line 1: I might suggest splitting each operation in its own file and putting them inside a folder for reducing the size of this.

arwen/elrondapi/bigFloatOps.go
andrei-diaconescu: if op2 is zero it won't panic. you'll get +- infinity . Then the next operation you will want to use that value, you will get execution failed as no operation with infinity is allowed. However, I will add extra checks everywhere so you cannot even initialize big floats  with +- infinity.

arwen/elrondapi/bigFloatOps.go
andrei-diaconescu: yes, that's why I check if op.IsInt() then return and no Sub() is done

arwen/elrondapi/bigFloatOps.go
andrei-diaconescu: same as floor

arwenmandos/gasSchedules/gasScheduleV3.toml
```diff
@@ -166,6 +166,26 @@
...
+[BigFloatAPICost]
+    BigFloatNew         = 2000
+    BigFloatAdd         = 2000
+    BigFloatSub         = 2000
+    BigFloatMul         = 2000
+    BigFloatDiv         = 2000
+    BigFloatTruncate    = 2000
+    BigFloatMod         = 2000
+    BigFloatNeg         = 2000
+    BigFloatClone        = 2000
```
sasurobert, line 178: indent

config/config.toml
```diff
@@ -121,6 +121,26 @@
...
+[BigFloatAPICost]
+    BigFloatNew         = 10
+    BigFloatAdd         = 10
+    BigFloatSub         = 10
+    BigFloatMul         = 10    
+    BigFloatDiv         = 10
+    BigFloatTruncate    = 10
+    BigFloatMod         = 10
+    BigFloatNeg         = 10
+    BigFloatClone        = 10
```
sasurobert, line 133: indent

arwen/elrondapi/bigFloatOps.go
sasurobert: no splitting these are a lot of common things use here and there. common functions

arwen/elrondapi/bigFloatOps.go
sasurobert: float64 operations are deterministic

arwen/elrondapi/manBufOps.go
```diff
@@ -424,6 +432,59 @@ func v1_4_mBufferFromBigIntSigned(context unsafe.Pointer, mBufferHandle int32, b
...
+
+	gasToUse := metering.GasSchedule().ManagedBufferAPICost.MBufferFromBigFloat
+	metering.UseGas(gasToUse)
+
+	value, err := managedType.GetBigFloat(bigFloatHandle)
+	if arwen.WithFault(err, context, runtime.BigFloatAPIErrorShouldFailExecution()) {
+		return 1
+	}
+
+	encodedFloat, err := value.GobEncode()
```
andrei-marinica, line 478: Is this ok with +/- inf, or should be add a check?

andrei-marinica (changes requested):

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,699 @@
...
+	gasToUse := metering.GasSchedule().BigFloatAPICost.BigFloatNeg
+	metering.UseGas(gasToUse)
+
+	dest := managedType.GetBigFloatOrCreate(destinationHandle)
+	op, err := managedType.GetBigFloat(opHandle)
+	if arwen.WithFault(err, context, runtime.BigFloatAPIErrorShouldFailExecution()) {
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(dest, op)
+	dest.Neg(op)
```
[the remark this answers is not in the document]
andrei-diaconescu: you shouldn't be able to set an infinity in the big.Floats map, so Neg() couldn't possibly set an infinity too

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,699 @@
...
+
+	gasToUse := metering.GasSchedule().BigFloatAPICost.BigFloatAbs
+	metering.UseGas(gasToUse)
+
+	dest := managedType.GetBigFloatOrCreate(destinationHandle)
+	op, err := managedType.GetBigFloat(opHandle)
+	if arwen.WithFault(err, context, runtime.BigFloatAPIErrorShouldFailExecution()) {
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(dest, op)
```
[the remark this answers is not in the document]
andrei-diaconescu: same thing here. shouldn't be able to have an infinity from the big.Float map

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,699 @@
...
+	}
+	if oneIsInfinity(op) {
+		_ = arwen.WithFault(arwen.ErrInfinityFloatOperation, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return -1
+	}
+	bigIntOp := new(big.Int)
+	op.Int(bigIntOp)
+	managedType.ConsumeGasForBigFloatCopy(op)
+	managedType.ConsumeGasForBigIntCopy(bigIntOp)
+	return int32(bigIntOp.BitLen() - 1)
```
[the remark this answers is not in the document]
andrei-diaconescu: deleted

arwen/elrondapi/manBufOps.go
andrei-diaconescu: no need to check

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,650 @@
...
+	if op.IsInt() {
+		return
+	}
+	if bigIntOp.Sign() < 0 {
+		bigIntOp.Sub(bigIntOp, big.NewInt(1))
+	}
+}
+
+//export v1_4_bigFloatCeil
+func v1_4_bigFloatCeil(context unsafe.Pointer, opHandle, bigIntHandle int32) {
```
camilbancioiu, line 492: Consider renaming `bigIntHandle` to `destBigIntHandle`.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,650 @@
...
+	if op.IsInt() {
+		return
+	}
+	if bigIntOp.Sign() > 0 {
+		bigIntOp.Add(bigIntOp, big.NewInt(1))
+	}
+}
+
+//export v1_4_bigFloatTruncate
+func v1_4_bigFloatTruncate(context unsafe.Pointer, opHandle, bigIntHandle int32) {
```
camilbancioiu, line 517: Consider renaming `bigIntHandle` to `destBigIntHandle`.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,650 @@
...
+	metering.UseGas(gasToUse)
+
+	dest, err1 := managedType.GetBigFloatOrCreate(destinationHandle)
+	bigIntValue, err2 := managedType.GetBigInt(bigIntHandle)
+	if arwen.WithMultipleFaults([]error{err1, err2}, context, runtime.BigFloatAPIErrorShouldFailExecution()) {
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(dest)
+	managedType.ConsumeGasForBigIntCopy(bigIntValue)
+	dest.SetInt(bigIntValue)
```
camilbancioiu, line 586: Can we end up with an infinity here, due to representation issues? Big integers can get much bigger than big floats.

Consider creating an intermediary result, just like you did for the arithmetic operations, then check for infinity.

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,650 @@
...
+}
+
+//export v1_4_bigFloatMod
+func v1_4_bigFloatMod(context unsafe.Pointer, destinationHandle, op1Handle, op2Handle int32) {
```
camilbancioiu, line 590: Floating point modulo is not yet as well-defined in programming as other operations... Hopefully this won't be an issue. To be on the safe side, I would suggest disabling this API function until we clarify it.

https://en.wikipedia.org/wiki/Modulo_operation

arwen/elrondapi/manBufOps.go
camilbancioiu: This would be one of the few places where infinity won't cause issues :)

arwen/elrondapi/manBufOps.go
```diff
@@ -424,6 +432,67 @@ func v1_4_mBufferFromBigIntSigned(context unsafe.Pointer, mBufferHandle int32, b
...
+	if arwen.WithFault(err, context, runtime.ManagedBufferAPIErrorShouldFailExecution()) {
+		return 1
+	}
+
+	bigFloat := new(big.Float)
+	err = bigFloat.GobDecode(managedBuffer)
+	if arwen.WithFault(err, context, runtime.ManagedBufferAPIErrorShouldFailExecution()) {
+		return 1
+	}
+	if bigFloat.Prec() != 53 {
```
camilbancioiu, line 459: Use the constant `bigFloatPrecision` instead of `53`.

camilbancioiu (changes requested):

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,691 @@
...
+
+	imports, err = imports.Append("bigFloatGetConstE", v1_4_bigFloatGetConstE, C.v1_4_bigFloatGetConstE)
+	if err != nil {
+		return nil, err
+	}
+
+	return imports, err
+}
+
+func oneIsInfinity(values ...*big.Float) bool {
```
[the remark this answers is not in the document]
andrei-diaconescu: this function is no longer existent

arwen/elrondapi/bigFloatOps.go
andrei-diaconescu: https://pkg.go.dev/math#Mod

arwen/elrondapi/manBufOps.go
andrei-diaconescu: yep, anyways you absolutely shouldn't be able to get an infinity as if so, GetBigFloat() should return error and so fail execution

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,691 @@
...
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(dest, op1, op2)
+
+	resultAdd := new(big.Float).Add(op1, op2)
+	if oneIsInfinity(resultAdd) {
+		_ = arwen.WithFault(arwen.ErrInfinityFloatOperation, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(resultAdd)
```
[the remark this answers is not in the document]
andrei-diaconescu: actually no. good point. will delete

arwen/elrondapi/bigFloatOps.go
```diff
@@ -0,0 +1,691 @@
...
+	dest := managedType.GetBigFloatOrCreate(destinationHandle)
+	op1, op2, err := managedType.GetTwoBigFloat(op1Handle, op2Handle)
+	if arwen.WithFault(err, context, runtime.BigFloatAPIErrorShouldFailExecution()) {
+		return
+	}
+	if oneIsInfinity(op1, op2) {
+		_ = arwen.WithFault(arwen.ErrInfinityFloatOperation, context, runtime.BigFloatAPIErrorShouldFailExecution())
+		return
+	}
+	managedType.ConsumeGasForBigFloatCopy(dest, op1, op2)
```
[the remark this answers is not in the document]
andrei-diaconescu: it makes no difference but will change anyways

PR #398, opened 2021-08-27
Latest vm-common v1.1.6

iulianpascalau:
- latest vm-common v1.1.6


sasurobert:
new vmcommon was integrated in another pull request

PR #412, opened 2021-09-17
Wrapper for big float ops to avoid panics

andrei-diaconescu:
- there is still no limitation on how bigger a big Float can be. this will most probably be set depending on the exponent. This will come in a future PR with possible minor touches (like a gasSchedule) after benchmarks are done.


arwen/contexts/managedType.go
```diff
@@ -297,15 +297,15 @@ func (context *managedTypesContext) GetTwoBigFloats(handle1 int32, handle2 int32
 }
 
 // PutBigFloat adds the given value to the current values map and returns the handle
-func (context *managedTypesContext) PutBigFloat(value float64) int32 {
+func (context *managedTypesContext) PutBigFloat(value *big.Float) int32 {
```
iulianpascalau, line 300: please add a test that calls this function like:
```GO
//....
managedTypesContext.PutBigFloat(nil)
//...
```

Then add the required fixes.

math/bigFloat_wrapper_test.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func TestBigFloatSub_Panic(t *testing.T) {
+	value1 := new(big.Float)
+	_ = value1.GobDecode([]byte{1, 10, 0, 0, 0, 53, 0, 0, 0, 58, 0, 31, 28, 26, 150, 254, 14, 45})
+	value2 := new(big.Float)
+	_ = value2.GobDecode([]byte{1, 10, 0, 0, 0, 53, 0, 0, 0, 52, 222, 212, 49, 108, 64, 122, 107, 100})
+	result, err := SubBigFloat(value1, value2)
+
+	require.Equal(t, big.NewFloat(0).String(), result.String())
```
iulianpascalau, line 18: as discussed, this can be replaced to  
```
require.Equal(t, big.NewFloat(0), result)
```

if the defer functions rewrite the result pointer

iulianpascalau (approved):

math/overflow_test.go
```diff
@@ -17,8 +17,10 @@ func TestAddUint64(t *testing.T) {
 	c := uint64(3)
 	sum, err = AddUint64WithErr(b, c)
 	require.Equal(t, ErrAdditionOverflow, err)
+	require.Equal(t, b+c, sum)
```
camilbancioiu, line 20: Why should `sum` be equal to `b + c` (the normal golang sum) after `AddUint64WithErr` has already returned an overflow error?

math/overflow_test.go
```diff
@@ -17,8 +17,10 @@ func TestAddUint64(t *testing.T) {
...
 
 	d := uint64(0)
 	sum, err = AddUint64WithErr(b, d)
 	require.Nil(t, err)
+	require.Equal(t, b+d, sum)
```
camilbancioiu, line 25: Using `+` does not treat overflowing / underflowing. Please specify a value here, like `uint64(math.MaxUint64 - 2)` or whatever is expected.

test/contracts/big-floats/big-floats.c
```diff
@@ -147,135 +134,118 @@ void BigFloatCmpTest() {
...
         result = bigFloatSign(bigFloatHandle);
     }
     int64finish(result);
 }
 
 void BigFloatCloneTest() {
-    int reps, bigFloatHandle, result;
-    reps = int64getArgument(0);
-    bigFloatHandle = bigFloatNewFromFrac(0,1);
+    int result;
```
camilbancioiu, line 148: The variable `result` doesn't seem to be used in this function.

math/overflow_test.go
andrei-diaconescu: this is what the function returns.
`func AddUint64WithErr(a, b uint64) (uint64, error) {
	s := a + b
	if s >= a && s >= b {
		return s, nil
	}

	return s, ErrAdditionOverflow
}`

math/overflow_test.go
andrei-diaconescu: this is what the function returns.

```
func AddUint64WithErr(a, b uint64) (uint64, error) {
	s := a + b
	if s >= a && s >= b {
		return s, nil
	}

	return s, ErrAdditionOverflow
}
```

arwen/elrondapi/bigFloatOps.go
```diff
@@ -481,21 +521,29 @@ func v1_4_bigFloatPow(context unsafe.Pointer, destinationHandle, opHandle, expon
...
 
-func pow(base *big.Float, exp int32) *big.Float {
+func pow(context unsafe.Pointer, base *big.Float, exp int32) (*big.Float, error) {
 	result := big.NewFloat(1)
 	result.SetPrec(base.Prec())
+	runtime := arwen.GetRuntimeContext(context)
 	for i := 0; i < int(exp); i++ {
-		result.Mul(result, base)
+		resultMul, err := arwenMath.MulBigFloat(result, base)
+		if arwen.WithFault(err, context, runtime.BigFloatAPIErrorShouldFailExecution()) {
```
camilbancioiu, line 537: Is this `WithFault` necessary here? Could it be just `if err != nil`?

andrei-diaconescu: yes, the error returned from `arwenMath.MulBigFloat(result, base)` is basically the panic handled by the defer-recover mechanism.

test/contracts/big-floats/big-floats.c
andrei-diaconescu: this has been solved

camilbancioiu (approved):

arwen/elrondapi/bigFloatOps.go
```diff
@@ -543,7 +542,7 @@ func pow(context unsafe.Pointer, base *big.Float, exp int32) (*big.Float, error)
 }
 
 //export v1_4_bigFloatFloor
-func v1_4_bigFloatFloor(context unsafe.Pointer, opHandle, destBigIntHandle int32) {
+func v1_4_bigFloatFloor(context unsafe.Pointer, destBigIntHandle, opHandle int32) {
```
iulianpascalau, line 545: Using consecutive arguments of the same type is always tricky:
`func v1_4_bigFloatFloor(context unsafe.Pointer, destBigIntHandle, opHandle int32)`

This is prone to the following programming error:
v1_4_bigFloatFloor(context, opHandle, destBigIntHandle)` 
which the compiler does not catch (in many cases not even the reviewers)

iulianpascalau (approved):

PR #413, opened 2021-09-20
Gas tracing

andrei-diaconescu:


arwen/contexts/runtime.go
```diff
@@ -833,6 +838,31 @@ func (context *runtimeContext) GetAllErrors() error {
...
+	context.traceGasEnabled = true
+}
+
+// DisableTraceGas sets true to the flag variable that traces gas consumption
+func (context *runtimeContext) DisableTraceGas() {
+	context.traceGasEnabled = false
+}
+
+// TraceGasUsed adds the usedGas passed to the traced gas map with the key as the passed functionName
+func (context *runtimeContext) TraceGasUsed(functionName string, usedGas uint64) {
```
andrei-marinica, line 852: The runtime context has access to the current gas left. Could we also create a method that only takes the initial gas and computes the delta itself? So we don't have to write so much in each EI function?

arwenmandos/execScenario.go
```diff
@@ -32,6 +33,18 @@ func (ae *ArwenTestExecutor) ExecuteScenario(scenario *mj.Scenario, fileResolver
...
+	if scenario.TraceGas {
+		runtime := host.Runtime()
+		gasUsed := runtime.GetTracedGas()
+		for key, value := range gasUsed {
+			totalGasUsed := uint64(0)
+			for _, gasUsed := range value {
+				totalGasUsed += gasUsed
+			}
+			log.Trace("GasTrace", "functionName:", key, "totalGasUsed:", totalGasUsed)
+		}
```
andrei-marinica, line 45: Extract this block into a gas init method. Can be in the runtime context.

arwenmandos/execScenario.go
```diff
@@ -17,7 +17,8 @@ func (ae *ArwenTestExecutor) Reset() {
 func (ae *ArwenTestExecutor) ExecuteScenario(scenario *mj.Scenario, fileResolver fr.FileResolver) error {
 	ae.fileResolver = fileResolver
 	ae.checkGas = scenario.CheckGas
-	err := ae.InitVM(scenario.GasSchedule)
+	ae.traceGas = scenario.TraceGas
+	host, err := ae.InitVM(scenario.GasSchedule)
```
andrei-marinica, line 21: Don't return the host here, wrap related functionality in a method in the ArwenTestExecutor

mandos-go/json/write/writeScenario.go
```diff
@@ -28,6 +28,11 @@ func ScenarioToOrderedJSON(scenario *mj.Scenario) oj.OJsonObject {
 		scenarioOJ.Put("checkGas", &ojFalse)
 	}
 
+	if !scenario.TraceGas {
```
andrei-marinica, line 31: Make it default to be false. We don't want to trace the gas all the time.

mandos-go/json/write/writeTest.go
```diff
@@ -29,6 +29,11 @@ func testToOJ(test *mj.Test) oj.OJsonObject {
 		testOJ.Put("checkGas", &ojFalse)
 	}
 
+	if !test.TraceGas {
```
andrei-marinica, line 32: No need to add it here, these are legacy tests, will get removed at some point

mandos-go/json/model/structTest.go
```diff
@@ -4,6 +4,7 @@ package mandosjsonmodel
 type Test struct {
 	TestName    string
 	CheckGas    bool
+	TraceGas    bool
```
andrei-marinica, line 7: Not needed here, this is legacy code, no longer used.

andrei-marinica (changes requested):

arwen/elrondapi/elrondei.go
```diff
@@ -2442,6 +2713,13 @@ func v1_4_executeOnDestContext(
 	dataOffset int32,
 ) int32 {
 	host := arwen.GetVMHost(context)
+	runtime := host.Runtime()
+	metering := host.Metering()
+	initialGasLeft := metering.GasLeft()
+	defer func() {
```
andrei-marinica, line 2719: won't this include all the gas consumed within the called contract?
I guess it's fine for the first version, but we should think about this

andrei-marinica (approved):

arwen/contexts/runtime.go
```diff
@@ -833,6 +838,37 @@ func (context *runtimeContext) GetAllErrors() error {
...
 
+// EnableTraceGas sets true to the flag variable that traces gas consumption
+func (context *runtimeContext) EnableTraceGas() {
+	context.traceGasEnabled = true
+}
+
+// DisableTraceGas sets true to the flag variable that traces gas consumption
+func (context *runtimeContext) DisableTraceGas() {
+	context.traceGasEnabled = false
+}
```
camilbancioiu, line 849:
```suggestion
// EnableGasTrace sets true to the flag variable that traces gas consumption
func (context *runtimeContext) EnableGasTrace() {
	context.traceGasEnabled = true
}

// DisableGasTrace sets true to the flag variable that traces gas consumption
func (context *runtimeContext) DisableGasTrace() {
	context.traceGasEnabled = false
}
```

arwen/contexts/runtime.go
```diff
@@ -833,6 +838,37 @@ func (context *runtimeContext) GetAllErrors() error {
...
+
+func (context *runtimeContext) computeUsedGas(initialGasLeft uint64) uint64 {
+	gasLeft := context.host.Metering().GasLeft()
+	return initialGasLeft - gasLeft
+}
+
+// GetTracedGas returns the gasTrace map
+func (context *runtimeContext) GetTracedGas() map[string][]uint64 {
+	return context.gasTrace
+}
```
camilbancioiu, line 870:
```suggestion
// GetGasTrace returns the gasTrace map
func (context *runtimeContext) GetGasTrace() map[string][]uint64 {
	return context.gasTrace
}
```

arwenmandos/util.go
```diff
@@ -198,3 +199,21 @@ func addESDTToVMInput(esdtData []*mj.ESDTTxData, vmInput *vmcommon.VMInput) {
...
+	runtime := ae.GetVMHost().Runtime()
+	gasUsed := runtime.GetTracedGas()
+	arwen.SetLoggingForTests()
+	for key, value := range gasUsed {
+		totalGasUsed := uint64(0)
+		for _, gasUsed := range value {
+			totalGasUsed += gasUsed
+		}
+		log.Trace("GasTrace", "functionName:", key, ",  totalGasUsed:", totalGasUsed)
+	}
```
camilbancioiu, line 213:
```suggestion
	gasTrace := runtime.GetGasTrace()
	arwen.SetLoggingForTests()
	for key, value := range gasTrace {
		totalGasUsed := uint64(0)
		for _, gasUsed := range value {
			totalGasUsed += gasUsed
		}
		log.Trace("GasTrace", "functionName:", key, ",  totalGasUsed:", totalGasUsed)
	}
```

cmd/test/runner.go
```diff
@@ -64,7 +64,8 @@ func main() {
 			executor,
 			mc.NewDefaultFileResolver(),
 		)
-		err = runner.RunSingleJSONScenario(jsonFilePath)
+		traceGas := false
+		err = runner.RunSingleJSONScenario(jsonFilePath, traceGas)
```
camilbancioiu, line 68: Passing `traceGas` will override whatever the JSON scenario contains. Don't pass it at all.

arwen/interface.go
```diff
@@ -145,6 +145,10 @@ type RuntimeContext interface {
 	BigIntAPIErrorShouldFailExecution() bool
 	ManagedBufferAPIErrorShouldFailExecution() bool
 	ExecuteAsyncCall(address []byte, data []byte, value []byte) error
+	EnableTraceGas()
+	DisableTraceGas()
+	TraceGasUsed(functionName string, initialGasLeft uint64)
+	GetTracedGas() map[string][]uint64
```
camilbancioiu, line 151:
```suggestion
	EnableGasTrace()
	DisableGasTrace()
	TraceGasUsed(functionName string, initialGasLeft uint64)
	GetGasTrace() map[string][]uint64
```

camilbancioiu (dismissed):
Please remove gas tracing from `asyncCall`, `createAsyncCall`, `executeOnSameContext`,  `executeOnDestContext` and its variants, `upgradeContract` and its variants. These functions call other contracts and will distort the gas trace.

arwen/contexts/runtime.go
```diff
@@ -833,6 +838,37 @@ func (context *runtimeContext) GetAllErrors() error {
 	return context.errors
 }
 
+// EnableGasTrace sets true to the flag variable that traces gas consumption
```
sasurobert, line 841: init gas trace on InitState

arwen/contexts/runtime.go
```diff
@@ -833,6 +838,37 @@ func (context *runtimeContext) GetAllErrors() error {
...
+}
+
+func (context *runtimeContext) computeUsedGas(initialGasLeft uint64) uint64 {
+	gasLeft := context.host.Metering().GasLeft()
+	return initialGasLeft - gasLeft
+}
+
+// GetGasTrace returns the gasTrace map
+func (context *runtimeContext) GetGasTrace() map[string][]uint64 {
+	return context.gasTrace
```
sasurobert, line 869: make gasTrace as map[string]map[string][]uint64

arwen/elrondapi/manBufOps.go
```diff
@@ -152,6 +152,11 @@ func ManagedBufferImports(imports *wasmer.Imports) (*wasmer.Imports, error) {
 func v1_4_mBufferNew(context unsafe.Pointer) int32 {
 	managedType := arwen.GetManagedTypesContext(context)
 	metering := arwen.GetMeteringContext(context)
+	runtime := arwen.GetRuntimeContext(context)
+	initialGasLeft := metering.GasLeft()
+	defer func() {
```
sasurobert, line 157: No defer func please - this makes the code much slower

arwen/interface.go
```diff
@@ -145,6 +145,10 @@ type RuntimeContext interface {
 	BigIntAPIErrorShouldFailExecution() bool
 	ManagedBufferAPIErrorShouldFailExecution() bool
 	ExecuteAsyncCall(address []byte, data []byte, value []byte) error
+	EnableGasTrace()
+	DisableGasTrace()
+	TraceGasUsed(functionName string, initialGasLeft uint64)
+	GetGasTrace() map[string][]uint64
```
sasurobert, line 151: Define a struct for the gasTrace and return that

arwenmandos/exec.go
```diff
@@ -89,6 +93,10 @@ func (ae *ArwenTestExecutor) GetVM() vmi.VMExecutionHandler {
 	return ae.vm
 }
 
+func (ae *ArwenTestExecutor) GetVMHost() arwen.VMHost {
```
sasurobert, line 96: add a comment to the public function

mandos-go/controller/scenarioOne.go
```diff
@@ -35,6 +35,9 @@ func (r *ScenarioRunner) RunSingleJSONScenario(contextPath string) error {
 
 	r.Parser.ExprInterpreter.FileResolver.SetContext(contextPath)
 	scenario, parseErr := r.Parser.ParseScenarioFile(byteValue)
+	if len(traceGas) != 0 {
+		scenario.TraceGas = traceGas[0]
```
sasurobert, line 39: this seems strange ? why a list of bools when you use only one ?

mock/context/runtimeContextMock.go
```diff
@@ -348,3 +350,24 @@ func (r *RuntimeContextMock) AddError(err error, otherInfo ...string) {
 func (r *RuntimeContextMock) GetAllErrors() error {
 	return nil
 }
+
```
sasurobert, line 353: add comment to public functions

arwen/elrondapi/bigIntOps.go
```diff
@@ -1074,6 +1233,10 @@ func v1_4_bigIntFinishSigned(context unsafe.Pointer, referenceHandle int32) {
 	output := arwen.GetOutputContext(context)
 	metering := arwen.GetMeteringContext(context)
 	runtime := arwen.GetRuntimeContext(context)
+	initialGasLeft := metering.GasLeft()
+	defer func() {
+		runtime.TraceGasUsed("bigIntFinishSigned", initialGasLeft)
```
sasurobert, line 1238: define all these string as constants please

sasurobert (changes requested):

arwen/contexts/runtime.go
```diff
@@ -833,6 +838,37 @@ func (context *runtimeContext) GetAllErrors() error {
...
+	context.traceGasEnabled = false
+}
+
+// TraceGasUsed adds the usedGas passed to the traced gas map with the key as the passed functionName
+func (context *runtimeContext) TraceGasUsed(functionName string, initialGasLeft uint64) {
+	if context.traceGasEnabled {
+		if context.gasTrace[functionName] == nil {
+			context.gasTrace[functionName] = make([]uint64, 0)
+		}
+		usedGas := context.computeUsedGas(initialGasLeft)
```
sasurobert, line 857: It would be better to compute the total Gas and give that one here, instead initialGasLeft

sasurobert (changes requested):

arwen/contexts/runtime.go
andrei-diaconescu: the gasTracing printed would only be from the last runtime

arwen/elrondapi/bigIntOps.go
andrei-diaconescu: there would be over 100 string constants if I do that. I know it's because of the linter's warnings probably (mine doesn't show such warnings), but do we really want 100+ string constants?

sasurobert: yes. in terms of compilation and execution speed it will be much faster.

arwen/contexts/runtime.go
sasurobert: GasTracing should be per transaction execution. As it is done right now, this is a memory leak. Even in mandos we need to print gasTrace after evey step.

arwen/contexts/runtime.go
andrei-marinica: I asked him for this method because it creates less copy-paste code in the hooks.
We could also have another method that just does the trace, without computing the delta.

andrei-marinica: func (context *runtimeContext) TraceGas(functionName string, initialGasLeft uint64) { ... }

func (context *runtimeContext) TraceGasUsed(functionName string, initialGasLeft uint64) {
    context.TraceGas(context.computeUsedGas(initialGasLeft))
}

sasurobert: This is not yet resolved.

arwen/contexts/runtime.go
```diff
@@ -833,6 +846,69 @@ func (context *runtimeContext) GetAllErrors() error {
...
+	if context.currentGasTracer.gasTrace[scAddress][functionName] == nil {
+		context.currentGasTracer.gasTrace[scAddress][functionName] = make([]uint64, 0)
+	}
+}
+
+// ComputeAndSetUsedGasInGasTrace computes the used gas in the gasTrace map. This function should be used afte initial gas has been set in the gasTrace map using SetInitialGasInGasTrace.
+func (context *runtimeContext) ComputeAndSetUsedGasInGasTrace() {
+	if context.currentGasTracer.traceGasEnabled {
+		functionName := context.currentGasTracer.currentFunctionNameTraced
+		numberOfCalls := len(context.currentGasTracer.gasTrace[string(context.scAddress)][functionName])
```
sasurobert, line 897: this can pannic if there is nothing under context.scAddress or stuff like that. Need more protection

arwen/cryptoapi/cryptoei.go
```diff
@@ -197,10 +224,12 @@ func v1_4_keccak256(context unsafe.Pointer, dataOffset int32, length int32, resu
 	runtime := arwen.GetRuntimeContext(context)
 	crypto := arwen.GetCryptoContext(context)
 	metering := arwen.GetMeteringContext(context)
+	initialGasLeft := metering.GasLeft()
 
 	memLoadGas := math.MulUint64(metering.GasSchedule().BaseOperationCost.DataCopyPerByte, uint64(length))
 	gasToUse := math.AddUint64(metering.GasSchedule().CryptoAPICost.Keccak256, memLoadGas)
 	metering.UseGas(gasToUse)
+	runtime.TraceGasUsed(keccak256Name, initialGasLeft)
```
sasurobert, line 232: it would be pretty easy to do runtime.TraceGasUsed(keccak256name, gasToUse) instead of inital gas left and computation

arwenmandos/execScenario.go
```diff
@@ -17,18 +17,21 @@ func (ae *ArwenTestExecutor) Reset() {
...
+	resetGasTracesIfNewTest(ae, scenario)
+
 	err := ae.InitVM(scenario.GasSchedule)
 	if err != nil {
 		return err
 	}
 
 	txIndex := 0
 	for _, generalStep := range scenario.Steps {
+		enableGasTraceInRuntime(ae)
```
sasurobert, line 29: instead of enable and disable you could have one function only Set - which receives a bool. Simplifies this a lot.

sasurobert (changes requested):

arwen/elrondapi/managedei.go
```diff
@@ -252,14 +287,17 @@ func v1_4_managedWriteLog(
 	output := arwen.GetOutputContext(context)
 	metering := arwen.GetMeteringContext(context)
 	managedType := arwen.GetManagedTypesContext(context)
+	initialGasLeft := metering.GasLeft()
 
 	topics, sumOfTopicByteLengths, err := readManagedVecOfManagedBuffers(managedType, topicsHandle)
 	if arwen.WithFault(err, context, runtime.ElrondAPIErrorShouldFailExecution()) {
+		runtime.TraceGasUsed(managedWriteLogName, initialGasLeft)
```
sasurobert, line 294: no need to trace gas on ERROR

sasurobert (changes requested):

andrei-marinica (approved):

arwen/contexts/runtime.go
```diff
@@ -42,6 +44,12 @@ type runtimeContext struct {
 	errors          arwen.WrappableError
 }
 
+type gasTracer struct {
+	traceGasEnabled           bool
+	gasTrace                  map[string]map[string][]uint64
```
camilbancioiu, line 49: Consider defining a type alias for `map[string]map[string][]uint64`, e.g. a type called `GasTraceMap`.

arwen/cryptoapi/cryptoei.go
```diff
@@ -259,17 +290,20 @@ func v1_4_verifyBLS(
...
 	crypto := arwen.GetCryptoContext(context)
 	metering := arwen.GetMeteringContext(context)
+	initialGasLeft := metering.GasLeft()
 
 	gasToUse := metering.GasSchedule().CryptoAPICost.VerifyBLS
 	metering.UseGas(gasToUse)
 
 	key, err := runtime.MemLoad(keyOffset, blsPublicKeyLength)
 	if arwen.WithFault(err, context, runtime.CryptoAPIErrorShouldFailExecution()) {
+		runtime.TraceGasUsed(verifyBLSName, initialGasLeft)
```
camilbancioiu, line 300: No need to trace gas after `arwen.WithFault()`.

arwen/contexts/runtime_test.go
```diff
@@ -66,9 +66,67 @@ func TestNewRuntimeContext(t *testing.T) {
...
+	runtimeContext.TraceGasUsed("secondFunctionName", 3000)
+	require.Equal(t, 1, len(runtimeContext.currentGasTracer.gasTrace))
+	require.Equal(t, 1, len(runtimeContext.currentGasTracer.gasTrace[scAddress]["secondFunctionName"]))
+
+	runtimeContext.TraceGasUsed("calledFunctionName", 50000)
+	require.Equal(t, 1, len(runtimeContext.currentGasTracer.gasTrace))
+	require.Equal(t, 2, len(runtimeContext.currentGasTracer.gasTrace[scAddress]["calledFunctionName"]))
+	require.Equal(t, uint64(50000), runtimeContext.currentGasTracer.gasTrace[scAddress]["calledFunctionName"][1])
+
+	expectedGasTraceMap := map[string]map[string][]uint64(map[string]map[string][]uint64{"firstScAddress": {"calledFunctionName": []uint64{0xbb80, 0xc350}, "secondFunctionName": []uint64{0xbb8}}})
```
camilbancioiu, line 103: Type alias `GasTraceMap` is definitelly needed.

camilbancioiu:
```suggestion
	expectedGasTraceMap := map[string]map[string][]uint64{
		"firstScAddress": {
			"calledFunctionName": []uint64{0xbb80, 0xc350},
			"secondFunctionName": []uint64{0xbb8},
		},
	}

```

arwen/cryptoapi/cryptoei.go
```diff
@@ -49,6 +49,31 @@ const secp256k1UncompressedPublicKeyLength = 65
...
+	scalarMultECName                = "scalarMultEC"
+	marshalECName                   = "marshalEC"
+	unmarshalECName                 = "unmarshalEC"
+	marshalCompressedECName         = "marshalCompressedEC"
+	unmarshalCompressedECName       = "unmarshalCompressedEC"
+	generateKeyECName               = "generateKeyEC"
+	createECName                    = "createEC"
+	getCurveLengthECName            = "getCurveLengthEC"
+	getPrivKeyByteLengthECName      = "getPrivKeyByteLengthEC"
+	ellipticCurveGetValuesName      = "ellipticCurveGetValues"
```
camilbancioiu, line 74: As a future improvement, we could use these constants in import builder methods such as `CryptoImports`. Not for this PR.

arwenmandos/util.go
```diff
@@ -198,3 +199,58 @@ func addESDTToVMInput(esdtData []*mj.ESDTTxData, vmInput *vmcommon.VMInput) {
...
+			log.Trace("Gas Trace for: ", "SC Address", scAddress)
+			for functionName, value := range gasTrace {
+				totalGasUsed := uint64(0)
+				for _, usedGas := range value {
+					totalGasUsed += usedGas
+				}
+				log.Trace("GasTrace", "functionName:", functionName, ",  totalGasUsed:", totalGasUsed, ", numberOfCalls:", len(value))
+				totalGasUsedInSmartContract += int(totalGasUsed)
+			}
+			log.Trace("TotalGasUsedInSmartContract", "", totalGasUsedInSmartContract)
```
camilbancioiu, line 220: Total gas used by the contract is `GasProvided` minus `GasRemaining` at the end. Instead, the value computed here is the total gas used by the contract _for API calls_. It's not all the gas.

arwen/contexts/runtime.go
```diff
@@ -42,6 +44,14 @@ type runtimeContext struct {
 	errors          arwen.WrappableError
 }
 
+type GasTraceMap map[string]map[string][]uint64
```
sasurobert, line 47: add comment for public type

arwen/contexts/runtime.go
```diff
@@ -833,6 +848,74 @@ func (context *runtimeContext) GetAllErrors() error {
...
+		context.createGasTraceIfNil(scAddress, functionName)
+		context.currentGasTracer.gasTrace[scAddress][functionName] = append(context.currentGasTracer.gasTrace[scAddress][functionName], usedGas)
+	}
+}
+
+// UseAndTraceGas uses the passed gasToUse and traces it
+func (context *runtimeContext) UseAndTraceGas(functionName string, gasToUse uint64) {
+	metering := context.host.Metering()
+	metering.UseGas(gasToUse)
+	if context.currentGasTracer.traceGasEnabled && context.scAddress != nil {
```
sasurobert, line 874: call TraceGasUsed instead of duplicated code here

arwen/contexts/runtime.go
```diff
@@ -833,6 +848,74 @@ func (context *runtimeContext) GetAllErrors() error {
...
+	}
+	if context.currentGasTracer.gasTrace[scAddress][functionName] == nil {
+		context.currentGasTracer.gasTrace[scAddress][functionName] = make([]uint64, 0)
+	}
+}
+
+// ComputeAndSetUsedGasInGasTrace computes the used gas in the gasTrace map. This function should be used afte initial gas has been set in the gasTrace map using SetInitialGasInGasTrace.
+func (context *runtimeContext) ComputeAndSetUsedGasInGasTrace() {
+	if context.currentGasTracer.traceGasEnabled && context.scAddress != nil {
+		functionName := context.currentGasTracer.currentFunctionNameTraced
```
sasurobert, line 903: current function name should be equal with context.VMInput.Function

arwen/contexts/runtime.go
```diff
@@ -833,6 +848,74 @@ func (context *runtimeContext) GetAllErrors() error {
...
+		numberOfCalls := len(context.currentGasTracer.gasTrace[string(context.scAddress)][functionName])
+
+		gasLeft := context.host.Metering().GasLeft()
+		usedGas := context.currentGasTracer.gasTrace[string(context.scAddress)][functionName][numberOfCalls-1] - gasLeft
+		context.currentGasTracer.gasTrace[string(context.scAddress)][functionName][numberOfCalls-1] = usedGas
+
+		context.currentGasTracer.currentFunctionNameTraced = ""
```
sasurobert, line 910: why do you rewrite this ?

arwen/contexts/runtime.go
```diff
@@ -833,6 +848,74 @@ func (context *runtimeContext) GetAllErrors() error {
...
+
+// ComputeAndSetUsedGasInGasTrace computes the used gas in the gasTrace map. This function should be used afte initial gas has been set in the gasTrace map using SetInitialGasInGasTrace.
+func (context *runtimeContext) ComputeAndSetUsedGasInGasTrace() {
+	if context.currentGasTracer.traceGasEnabled && context.scAddress != nil {
+		functionName := context.currentGasTracer.currentFunctionNameTraced
+		numberOfCalls := len(context.currentGasTracer.gasTrace[string(context.scAddress)][functionName])
+
+		gasLeft := context.host.Metering().GasLeft()
+		usedGas := context.currentGasTracer.gasTrace[string(context.scAddress)][functionName][numberOfCalls-1] - gasLeft
+		context.currentGasTracer.gasTrace[string(context.scAddress)][functionName][numberOfCalls-1] = usedGas
```
sasurobert, line 908: Please add to a new field this total gas used. why are you accumulating it to the end of the list.

sasurobert: It would be much better to define a set of functions for the gasTrace and move all the logic there. It does not look good to have gasTrace[string(context.scAddress)][functionName][numberOfCalls-1]

arwen/cryptoapi/cryptoei.go
```diff
@@ -344,6 +374,7 @@ func v1_4_verifyCustomSecp256k1(
 	runtime := arwen.GetRuntimeContext(context)
 	crypto := arwen.GetCryptoContext(context)
 	metering := arwen.GetMeteringContext(context)
+	runtime.SetInitialGasInGasTrace(verifyCustomSecp256k1Name)
```
sasurobert, line 377: please do not do this. There are only 2 UseGas in this function and you can simply use runtime.TraceGas(gasToUse)

arwen/cryptoapi/cryptoei.go
```diff
@@ -397,6 +429,9 @@ func v1_4_verifySecp256k1(
 	messageLength int32,
 	sigOffset int32,
 ) int32 {
+	runtime := arwen.GetRuntimeContext(context)
+	runtime.SetInitialGasInGasTrace(verifySecp256k1Name)
```
sasurobert, line 433: delete this function please. SetInitialGasInGasTrace(verifySecp256k1Name)

arwen/elrondapi/manBufOps.go
```diff
@@ -248,6 +274,7 @@ func v1_4_mBufferGetByteSlice(context unsafe.Pointer, sourceHandle int32, starti
 
 	gasToUse = math.MulUint64(metering.GasSchedule().BaseOperationCost.DataCopyPerByte, uint64(len(sourceBytes)))
 	metering.UseGas(gasToUse)
+	runtime.TraceGasUsed(mBufferGetByteSliceName, initialGasLeft-metering.GasLeft())
```
sasurobert, line 277: why do you need initialGasLeft - metering.GasLeft - here you know the actual used gas by doing a single SUM.

sasurobert: delete from all the places.

arwen/elrondapi/manBufOps.go
```diff
@@ -380,6 +416,7 @@ func v1_4_mBufferAppendBytes(context unsafe.Pointer, accumulatorHandle int32, da
 
 	gasToUse = math.MulUint64(metering.GasSchedule().BaseOperationCost.DataCopyPerByte, uint64(len(data)))
 	metering.UseGas(gasToUse)
+	runtime.TraceGasUsed(mBufferAppendBytesName, initialGasLeft-metering.GasLeft())
```
sasurobert, line 419: it would be better to have something like AddToTracedGas to simplify the actions.

arwen/elrondapi/managedei.go
```diff
@@ -708,6 +755,7 @@ func v1_4_managedExecuteReadOnly(
 		vmInput.arguments,
 	)
 	setReturnDataIfExists(host, lenReturnData, resultHandle)
+	runtime.TraceGasUsed(managedExecuteReadOnlyName, initialGasLeft-metering.GasLeft())
```
sasurobert, line 758: we actually want to trace the API call - not necessarily the whole execution below it. You already have that cost in the underlying execution. Here you will have duplicated stuffs as gas. You will have the same gas consumed twice in gasTrace

arwen/interface.go
```diff
@@ -145,6 +145,13 @@ type RuntimeContext interface {
 	BigIntAPIErrorShouldFailExecution() bool
 	ManagedBufferAPIErrorShouldFailExecution() bool
 	ExecuteAsyncCall(address []byte, data []byte, value []byte) error
+	EnableGasTrace()
+	DisableGasTrace()
+	TraceGasUsed(functionName string, initialGasLeft uint64)
```
sasurobert, line 150: we could have the Gas tracing as a separate component - outside of runtime context. It would be better.

arwen/elrondapi/bigIntOps.go
```diff
@@ -372,15 +416,21 @@ func v1_4_bigIntGetCallValue(context unsafe.Pointer, destinationHandle int32) {
...
 }
 
 //export v1_4_bigIntGetESDTCallValue
 func v1_4_bigIntGetESDTCallValue(context unsafe.Pointer, destination int32) {
-	_ = failIfMoreThanOneESDTTransfer(context)
+	runtime := arwen.GetRuntimeContext(context)
+	runtime.SetInitialGasInGasTrace(bigIntGetESDTCallValueName)
+	isFail := failIfMoreThanOneESDTTransfer(context)
+	if isFail {
+		runtime.ComputeAndSetUsedGasInGasTrace()
```
sasurobert, line 431: no need to compute if fail ...

arwen/elrondapi/elrondei.go
```diff
@@ -1942,17 +2068,25 @@ func v1_4_callValue(context unsafe.Pointer, resultOffset int32) int32 {
 
 //export v1_4_getESDTValue
 func v1_4_getESDTValue(context unsafe.Pointer, resultOffset int32) int32 {
-	_ = failIfMoreThanOneESDTTransfer(context)
+	runtime := arwen.GetRuntimeContext(context)
+	runtime.SetInitialGasInGasTrace(getESDTValueName)
+	isFail := failIfMoreThanOneESDTTransfer(context)
+	if isFail {
+		runtime.ComputeAndSetUsedGasInGasTrace()
```
sasurobert, line 2075: no need to compute if FAILED.

arwen/elrondapi/elrondei.go
```diff
@@ -1942,17 +2068,25 @@ func v1_4_callValue(context unsafe.Pointer, resultOffset int32) int32 {
 
 //export v1_4_getESDTValue
 func v1_4_getESDTValue(context unsafe.Pointer, resultOffset int32) int32 {
-	_ = failIfMoreThanOneESDTTransfer(context)
+	runtime := arwen.GetRuntimeContext(context)
+	runtime.SetInitialGasInGasTrace(getESDTValueName)
```
sasurobert, line 2072: delete set initial gas in gas trace function

sasurobert (changes requested):

arwen/elrondapi/elrondei.go
andrei-diaconescu: so then if you call `getESDTValue()` your gas would be traced under `getESDTValueByIndex`, it would have been more specific, but I will change

andrei-diaconescu: will do the same for the rest of pair functions like `functionName`, `functionNameByIndex`

arwen/contexts/runtime.go
andrei-diaconescu: because if there is another API function called which uses `SetInitialGasInGasTrace()` and we don't rewrite `""`, then it would be able to trace gas

arwen/contexts/runtime.go
andrei-diaconescu: the new field was added during call of `SetInitialGasInGasTrace()` , in this `ComputeAndSetUsedGasInGasTrace()` we only rewrite the initialGas set in the new field with the usedGas.

arwen/contexts/runtime.go
sasurobert: We should not use construction like this. Function SetInitialGasInGasTrace should not exist

arwen/contexts/runtime.go
sasurobert: Function SetInitialGasInGasTrace should not exist

arwen/elrondapi/elrondei.go
sasurobert: that is fair

arwen/contexts/gasTracer.go
```diff
@@ -0,0 +1,88 @@
+package contexts
+
+// GasTraceMap is the map that holds gas traces for all API functions, depending on SCAddress and functionName
+type gasTraceMap map[string]map[string][]uint64
+
+// GasTracer holds information about the gas trace
+type GasTracer struct {
```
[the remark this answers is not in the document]
andrei-diaconescu: i created the GasTracing interface in `arwen` package

sasurobert (approved):

andrei-marinica (approved):

sasurobert (approved):

PR #416, opened 2021-09-23
Fix ESDTTransferOnCallback in-shard for multi-transfers

camilbancioiu:


arwen/mock/epochNotifierStub.go
```diff
@@ -0,0 +1,28 @@
...
+		if !check.IfNil(handler) {
+			handler.EpochConfirmed(0, 0)
+		}
+	}
+}
+
+// IsInterfaceNil -
+func (ens *EpochNotifierStub) IsInterfaceNil() bool {
+	return ens == nil
+}
```
dorin-iancu, line 28: Won't this give a nil pointer exception if ens is nil? (i.e. `ens.IsInterfaceNil()`)

sasurobert: we are using check.IfNil function which is doing if object == nil || object.IsInterfaceNil return true

dorin-iancu (approved):

sasurobert (approved):

PR #423, opened 2021-10-06
Mandos go display logs opt

catalinnnn:


arwen/helpers.go
```diff
@@ -104,6 +104,11 @@ func SetLoggingForTests() {
 	logger.ToggleLoggerName(true)
 }
 
+// DisableLoggingForTests sets log level to *:NONE
+func DisableLoggingForTests() {
+	_ = logger.SetLogLevel("*:NONE")
```
andrei-marinica, line 109: I would go one step further and let mandos also specify the log level "NONE", "INFO", "DEBUG" ... I think

mandos-go/json/parse/parseScenario.go
```diff
@@ -267,6 +267,8 @@ func (p *Parser) parseTxStep(txType mj.TransactionType, stepMap *oj.OJsonMap) (*
 			if err != nil {
 				return nil, fmt.Errorf("bad tx step id: %w", err)
 			}
+		case "displayLogs":
```
andrei-marinica, line 270: Need to also add to the writer.
Also add an example in `examples.scen.json`. This way we can test that parsing works correctly.

mandos-go/json/parse/parseValue.go
```diff
@@ -141,6 +141,14 @@ func (p *Parser) parseString(obj oj.OJsonObject) (string, error) {
 	return str.Value, nil
 }
 
+func (p *Parser) parseBoolOrFalse(obj oj.OJsonObject) bool {
```
andrei-marinica, line 144: The preferred pattern is to return an error not a bool.

andrei-marinica (changes requested):

arwen/helpers.go
catalinnnn: This should be easy to use. Logs = true or false is what we need.

mandos-go/json/parse/parseScenario.go
catalinnnn: I added a test. I don't think printing the bool in writer is relevant

mandos-go/json/integrationTests/example.scen.json
```diff
@@ -257,6 +257,7 @@
             "step": "scCall",
             "txId": "1c",
             "comment": "without logs",
+            "displayLogs": true,
```
dorin-iancu, line 260: comment: without logs
displayLogs: true 

:thinking:

mandos-go/json/parse/parseValue.go
andrei-marinica: It will return false if you type "displayLogs": "true" for instance, instead of an error.
For when it's missing, the default is somewhere else.

mandos-go/json/integrationTests/example.scen.json
catalinnnn: not my fault that events are translated to logs.

catalinnnn: moved test to other step

andrei-marinica (approved):

PR #430, opened 2021-10-12
Fix OutOfGas hidden behind ExecutionFailed

camilbancioiu:
Due to a recent change, certain `OutOfGas` situations were hidden behind `ExecutionFailed`, which was incorrect. This PR restores the previous state.

Also added extra helper methods to the `VMOutputVerifier`.

### DIFF
--- a/arwen/contexts/runtime.go
+++ b/arwen/contexts/runtime.go
@@ -13,7 +13,6 @@ import (
 	"github.com/ElrondNetwork/arwen-wasm-vm/v1_4/wasmer"
 	logger "github.com/ElrondNetwork/elrond-go-logger"
 	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
-	"github.com/ElrondNetwork/elrond-vm-common/atomic"
 )
 
 var logRuntime = logger.GetOrCreate("arwen/runtime")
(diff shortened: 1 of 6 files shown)


arwen/contexts/runtime.go
```diff
@@ -466,16 +466,21 @@ func (context *runtimeContext) FailExecution(err error) {
...
+			breakpoint = arwen.BreakpointOutOfGas
+		}
 	} else {
 		message = "execution failed"
 		context.AddError(errors.New(message))
 	}
 
 	context.host.Output().SetReturnMessage(message)
-	context.SetRuntimeBreakpointValue(arwen.BreakpointExecutionFailed)
+	context.SetRuntimeBreakpointValue(breakpoint)
```
sasurobert, line 483: this is not backward compatible

sasurobert: add flag

camilbancioiu: Added epoch flag.

arwen/contexts/runtime.go
```diff
@@ -40,23 +41,27 @@ type runtimeContext struct {
...
+
+	flagFixOOGReturnCode *atomic.Flag
 }
 
 // NewRuntimeContext creates a new runtimeContext
 func NewRuntimeContext(
 	host arwen.VMHost,
 	vmType []byte,
 	builtInFuncContainer vmcommon.BuiltInFunctionContainer,
+	flagFixOOGReturnCode *atomic.Flag,
```
sasurobert, line 53: no. please do not give pointer of the flag.

sasurobert: make a pulbic function on arwen main context which returns if this flag is true or false.

camilbancioiu: Replaced with getter on `vmHost`.

PR #437, opened 2021-10-20
Merge master into feat promises final

camilbancioiu:


sasurobert (approved):

arwen/contexts/managedType.go
```diff
@@ -171,7 +171,7 @@ func (context *managedTypesContext) ConsumeGasForThisIntNumberOfBytes(byteLen in
 	gasToUse := uint64(0)
 	metering := context.host.Metering()
 	if byteLen > maxBigIntByteLenForNormalCost {
-		gasToUse = math.MulUint64(uint64(byteLen), metering.GasSchedule().BaseOperationCost.DataCopyPerByte)
+		gasToUse = math.MulUint64(uint64(byteLen), metering.GasSchedule().BigIntAPICost.CopyPerByteForTooBig)
```
andrei-diaconescu, line 174: I don't really get the logic of how big `CopyPerByteForTooBig` is. Why should 1000 gas cost be charged for numbers that use more than 32bytes, for each extra byte.
1.Say you've got 2 bigInt values of 32bytes, you do an Add() and that costs you the bigIntAdd GasCost which is 2k gas.
2.Then you've got 2 bigInt values of 33 bytes, you do an Add() and that costs you more than double(bigIntAdd + the extra gas you pay for the bigInt values that are "too big".

unfair and i think we could revise this "CopyPerByteForTooBig" cost, as imo it is too big

andrei-diaconescu (approved):

PR #474, opened 2021-11-25
execute on time context

sasurobert:
execute on time context


arwen/errors.go
```diff
@@ -23,6 +23,9 @@ var ErrSignalError = errors.New("error signalled by smartcontract")
 // ErrExecutionFailed signals that the execution failed
 var ErrExecutionFailed = errors.New("execution failed")
 
+// ErrExecutionFailedWithTimeout signals that the execution failed with timeout
+var ErrExecutionFailedWithTimeout = errors.New("execution failed")
```
bogdan-rosianu, line 27: might rename to `execution timed out` as discussed internally

arwen/host/arwen.go
```diff
@@ -311,69 +306,55 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
 	host.mutExecution.RLock()
 	defer host.mutExecution.RUnlock()
 
+	ctx, cancel := context.WithTimeout(context.Background(), time.Second*2)
```
bogdan-rosianu, line 309: might export a constant or even a config/flag for the 2 seconds value

arwen/host/arwen.go
```diff
@@ -311,69 +306,55 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
...
+	defer cancel()
+
 	log.Trace("RunSmartContractCreate begin", "len(code)", len(input.ContractCode), "metadata", input.ContractCodeMetadata)
 
-	try := func() {
+	done := make(chan interface{}, 1)
```
iulianpascalau, line 314:
```GO
done := make(chan struct{})
```
and on L317
```GO
close(done)
```

It is always safer to close the channel instead of writing a struct to it. Because someone can easily alter the code to unbuffer the chan, all tests pass but we end up with go routines leakage.

valid for all occurrences

arwen/host/arwen.go
iulianpascalau, line 309: maybe define as a const the time.Second*2 and use it also on 334

iulianpascalau (commented):
Although I'm a big fan of having a boxed-execution time frame for arwen to do its job I'm wondering if this mechanism couldn't block somehow a shard by creating a block that can never be processed with the same results by other nodes:
what happens if, due to a bug or mishap, the leader and the validators tried to execute the transaction and timed out? If the block can be successfully be proposed, this is a major issue.

arwen/host/arwen.go
```diff
@@ -311,69 +306,55 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
...
+	go func() {
 		vmOutput = host.doRunSmartContractCreate(input)
-	}
-
-	catch := func(caught error) {
-		err = caught
-		log.Error("RunSmartContractCreate", "error", err)
-	}
+		done <- 1
```
bogdan-rosianu, line 317: not sure if there is a case, but if it is, you could have a defer function inside this go routine that treats any panic that could have occurred (by using recover). this way, we can isolate entirely the sc call execution

sasurobert: yes, panic recovery can be added - although it won't happen :dagger:

arwen/errors.go
sasurobert: done

arwen/host/arwen.go
sasurobert: done

sasurobert: done

arwen/host/arwen.go
sasurobert: done

bogdan-rosianu (dismissed):

arwen/host/arwen.go
```diff
@@ -319,69 +316,55 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
...
+	}()
 
-	TryCatch(try, catch, "arwen.RunSmartContractCreate")
-	if vmOutput != nil {
-		log.Trace("RunSmartContractCreate end", "returnCode", vmOutput.ReturnCode, "returnMessage", vmOutput.ReturnMessage)
+	select {
+	case <-done:
+		return
+	case <-ctx.Done():
+		err = arwen.ErrExecutionFailedWithTimeout
```
camilbancioiu, line 334: It is important to also call `host.Runtime().FailExecution(err)` here, when the time is out before `done`. Otherwise the contract will be still running in the go-routine.

arwen/host/arwen.go
```diff
@@ -319,69 +316,55 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
...
+
+		close(done)
 	}()
 
-	try()
+	select {
+	case <-done:
+		return
+	case <-ctx.Done():
+		err = arwen.ErrExecutionFailedWithTimeout
```
camilbancioiu, line 365: Same as above, `host.Runtime().FailExecution(err)` is also needed.

camilbancioiu (changes requested):

arwen/host/arwen.go
```diff
@@ -319,69 +316,55 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
...
-			err, ok := r.(error)
-			if !ok {
-				err = fmt.Errorf("%s, panic: %v", catchFallbackMessage, r)
-			}
+	log.Trace("RunSmartContractCall begin", "function", input.Function)
 
-			catch(err)
+	done := make(chan struct{})
+	go func() {
+		isUpgrade := input.Function == arwen.UpgradeFunctionName
```
iulianpascalau, line 352: what happens, if, for some reason, this take too long, say, 10 seconds and the next call to the same contract, same function, same parameters comes to be executed?

`doRunSmartContractUpgrade`/`doRunSmartContractCall` will be ran in parallel. Valid also for `doRunSmartContractCreate`

Please add tests for the refactored functions

sasurobert: there is no parallel execution with the same VM.

iulianpascalau: if the `host.Runtime().FailExecution(err)` synchronously stop the execution, it is fine.

iulianpascalau (approved):

camilbancioiu (approved):

arwen/host/arwen.go
camilbancioiu: Yes, `FailExecution` injects an imminent breakpoint in Wasmer. The contract will not be able to continue afterwards.

arwen/host/arwen.go
```diff
@@ -319,69 +316,57 @@ func (host *vmHost) RunSmartContractCreate(input *vmcommon.ContractCreateInput)
...
-		TryCatch(tryCall, catch, "arwen.RunSmartContractCall")
-	}
-
-	return
-}
-
-// TryCatch simulates a try/catch block using golang's recover() functionality
-func TryCatch(try TryFunction, catch CatchFunction, catchFallbackMessage string) {
-	defer func() {
-		if r := recover(); r != nil {
```
iulianpascalau, line 374: I would have kept the recovery. Not a big overhead but a great protection when executing VM calls.

iulianpascalau (approved):

camilbancioiu (approved):

bogdan-rosianu (approved):

PR #478, opened 2021-11-26
Managed sha256 + keccak256 ei

andrei-marinica:


arwen/cryptoapi/cryptoei.go
```diff
@@ -217,6 +229,31 @@ func v1_4_sha256(context unsafe.Pointer, dataOffset int32, length int32, resultO
...
 }
 
+//export v1_4_managedSha256
+func v1_4_managedSha256(context unsafe.Pointer, inputHandle, outputHandle int32) int32 {
+	managedType := arwen.GetManagedTypesContext(context)
+	runtime := arwen.GetRuntimeContext(context)
+	crypto := arwen.GetCryptoContext(context)
+	metering := arwen.GetMeteringContext(context)
+
+	metering.UseGasAndAddTracedGas(keccak256Name, metering.GasSchedule().CryptoAPICost.SHA256)
```
camilbancioiu, line 239:
```suggestion
	metering.UseGasAndAddTracedGas(sha256Name, metering.GasSchedule().CryptoAPICost.SHA256)
```

camilbancioiu (changes requested):

camilbancioiu:
The new managed API methods appear to consume just as much gas as the unmanaged ones, even if they should be much faster. We could create a new cost `ManagedCopyDataPerByte`, or simply `ManagedSHA256` and `ManagedKeccak256` with smaller values.

andrei-marinica:
> The new managed API methods appear to consume just as much gas as the unmanaged ones, even if they should be much faster. We could create a new cost `ManagedCopyDataPerByte`, or simply `ManagedSHA256` and `ManagedKeccak256` with smaller values.

Fully agree, but it's more important to just have this functionality for now, even if it could be made cheaper.

sasurobert:
> The new managed API methods appear to consume just as much gas as the unmanaged ones, even if they should be much faster. We could create a new cost `ManagedCopyDataPerByte`, or simply `ManagedSHA256` and `ManagedKeccak256` with smaller values.

There is no need for API costs in case of crypto the actual crypto computation consumes the most of the time.

sasurobert (approved):

camilbancioiu (approved):

PR #483, opened 2021-12-07
Deploy / Update contracts from source tests

matei-p:


sasurobert (dismissed):

arwen/elrondapi/elrondei.go
```diff
@@ -1776,6 +1776,11 @@ func upgradeContract(
 		value,
 	)
 	logEEI.Trace("upgradeContract", "error", err)
+
+	storage := host.Storage()
+	if storage.IsUseDifferentGasCostFlagSet() && arwen.WithFaultAndHost(host, err, runtime.ElrondAPIErrorShouldFailExecution()) {
```
camilbancioiu, line 1781: I strongly recommend integrating the following condition:
```
	if errors.Is(err, arwen.ErrNotEnoughGas) {
		runtime.SetRuntimeBreakpointValue(arwen.BreakpointOutOfGas)
		return
	}
```

See `v1_4_asyncCall()` below, which does this.

arwen/hosttest/contracts_deploy_test.go
```diff
@@ -0,0 +1,466 @@
...
+	})
+}
+
+func TestDeployFromSource_NoGasForInit(t *testing.T) {
+	testConfig := getDeployFromSourceTestConfig()
+	testConfig.gasProvidedForInit = uint64(100)
+	runDeployFromSourceTest(t, &testConfig, func(world *worldmock.MockWorld, verify *test.VMOutputVerifier) {
+		verify.
+			ExecutionFailed().
+			HasRuntimeErrors(arwen.ErrInputAndOutputGasDoesNotMatch.Error())
```
camilbancioiu, line 98: So this doesn't fail with `OutOfGas`, but with `ExecutionFailed` due to gas mismatch? This is safe, not a problem, but must be investigated, this should not be a gas mismatch. Maybe leave a TODO?

camilbancioiu: I think integrating the condition I mentioned in `v1_4_upgradeContract()` will ensure the final error will be the proper `OutOfGas`. With the epoch flag, of course.

camilbancioiu (approved):

sasurobert (approved):

PR #492, opened 2022-01-15
Warm instances back

sasurobert:
Bring back warm instances - make them right.


arwen/contexts/runtime.go
```diff
@@ -206,6 +250,55 @@ func (context *runtimeContext) makeInstanceFromContractByteCode(contract []byte,
...
+		return false
+	}
+
+	warmInstanceMemory := warmInstance.GetMemory().Data()
+	if len(warmInstanceMemory) != len(localMemory) {
+		// TODO shrink the instance memory instead and return true
+		return false
+	}
+
+	copy(warmInstanceMemory, localMemory)
```
iulianpascalau, line 288: warmInstanceMemory (the destination on which the copy dumps the data) is not used in the rest of this function.

arwen/contexts/runtime.go
```diff
@@ -206,6 +250,55 @@ func (context *runtimeContext) makeInstanceFromContractByteCode(contract []byte,
...
+	}
+
+	localMemory, ok := cachedMemoryObject.([]byte)
+	if !ok {
+		return false
+	}
+
+	warmInstanceMemory := warmInstance.GetMemory().Data()
+	if len(warmInstanceMemory) != len(localMemory) {
+		// TODO shrink the instance memory instead and return true
```
iulianpascalau, line 284: why not always use the localMemory since that was cached last time? Why do we need the comparison with the "warmInstance" memory?

arwen/contexts/runtime.go
```diff
@@ -643,16 +752,9 @@ func (context *runtimeContext) GetInstanceExports() wasmer.ExportsMap {
 	return context.instance.GetExports()
 }
 
-// CleanWasmerInstance cleans the current wasmer instance.
+// CleanWasmerInstance cleans the current wasmer instance. it does nothing as cache and eviction are cleaning
```
iulianpascalau, line 755: Please remove this function.
We have `CleanWasmerInstance` and `ClearWarmInstanceCache`. Or make the CleanWasmerInstance contain the ClearWasmerInstance code. 
Otherwise this will confuse reviewers and maintainers of this code (felt in the the pitfall, thinking that you just used a useless function just about anywhere).

arwen/host/arwen.go
```diff
@@ -262,6 +262,11 @@ func (host *vmHost) InitState() {
 	host.initContexts()
 }
 
+// Close will close all underlying processes
+func (host *vmHost) Close() {
```
iulianpascalau, line 266: if you would have chosen to have the signature to something like
```GO
func (host *vmHost) Close() error
```
the arwen VM would have implemented `Closer` interface which is easier to integrate

go.mod
```diff
@@ -4,10 +4,10 @@ go 1.13
 
 require (
 	github.com/ElrondNetwork/big-int-util v0.1.0
-	github.com/ElrondNetwork/elrond-go-core v1.1.7
+	github.com/ElrondNetwork/elrond-go-core v1.1.8-0.20220112112600-5ad56d3f2305
```
iulianpascalau, line 7: do not forget about using proper releases

mandos-go/controller/testOne.go
```diff
@@ -27,7 +27,9 @@ func (r *TestRunner) RunSingleJSONTest(contextPath string) error {
 	}
 
 	// defer the closing of our jsonFile so that we can parse it later on
-	defer jsonFile.Close()
+	defer func() {
+		_ = jsonFile.Close()
```
iulianpascalau, line 31: :+1:

arwen/contexts/runtime.go
```diff
@@ -67,6 +74,16 @@ func NewRuntimeContext(
 		useDifferentGasCostForReadingCachedStorageEpoch: useDifferentGasCostForReadingCachedStorageEpoch,
 	}
 
+	var err error
+	context.warmInstanceCache, err = lrucache.NewCacheWithEviction(warmCacheSize, instanceEvicted)
```
iulianpascalau, line 78: I would have used only one cache containing both the instance pointer and the local data in a specialized, unexported struct. 
The reasons for this are the following:
1. in no way we could have the situation in which a local data value is evicted async with the instance (requiring the recreation of the warm instance). 
2. less type assertions as the only one used is between the cached object from the cache and the unexported struct containing both the local data and the instance pointer.

iulianpascalau (changes requested):
GJ! :+1:

arwen/contexts/runtime.go
sasurobert: but it is actually copied inside the memory space of the wasm code :dagger:

sasurobert: we need to restore memory always as it was from first compilation

arwen/contexts/runtime.go
sasurobert: we need to restore the memory as it was from first compilation. This means shrinking it if it used memory.grow last time.

arwen/host/arwen.go
sasurobert: ok :). will do. And it will be simpler in elrond-go.

arwen/contexts/runtime.go
sasurobert: good idea :). will do.

arwen/contexts/runtime.go
```diff
@@ -78,6 +95,19 @@ func NewRuntimeContext(
...
+	instance, ok := value.(wasmer.InstanceHandler)
+	if !ok {
+		return
+	}
+
+	instance.Clean()
+}
+
+func localDataEvicted(_ interface{}, value interface{}) {
+	value = nil
```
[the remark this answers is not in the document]
sasurobert: done

arwen/contexts/runtime.go
```diff
@@ -206,6 +250,55 @@ func (context *runtimeContext) makeInstanceFromContractByteCode(contract []byte,
 	return nil
 }
 
+func (context *runtimeContext) useWarmInstanceIfExists(gasLimit uint64, codeHash []byte, newCode bool) bool {
+	if newCode || len(codeHash) == 0 {
```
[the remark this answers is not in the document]
sasurobert: it helps in some context calls.

arwen/contexts/runtime.go
sasurobert: done

arwen/contexts/runtime.go
sasurobert: done

iulianpascalau (approved):
Looks better. 
Do not forget about proper releases
Still do not get the copy operation on runtime.go L273. `warmInstanceMemory` is not used anywhere

iulianpascalau (dismissed):

iulianpascalau (approved):

arwen/contexts/runtime.go
```diff
@@ -653,6 +746,7 @@ func (context *runtimeContext) CleanWasmerInstance() {
 	context.instance = nil
 
 	logRuntime.Trace("instance cleaned")
+	return
```
camilbancioiu, line 749: Forgotten `return`?

arwen/contexts/runtime.go
```diff
@@ -21,6 +23,8 @@ var logRuntime = logger.GetOrCreate("arwen/runtime")
 
 var _ arwen.RuntimeContext = (*runtimeContext)(nil)
 
+const warmCacheSize = 100
```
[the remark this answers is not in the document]
camilbancioiu: Optionally add a TODO with @iulianpascalau's suggestion.

camilbancioiu (approved):

PR #493, opened 2022-01-17
new vm-common on 1.3

sasurobert:
new vm-common

### DIFF
--- a/arwen/host/arwen.go
+++ b/arwen/host/arwen.go
@@ -382,6 +382,11 @@ func (host *vmHost) RunSmartContractCall(input *vmcommon.ContractCallInput) (vmO
 	return
 }
 
+// Close closes all internal instances of the vm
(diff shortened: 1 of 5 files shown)


arwen/host/arwen.go
```diff
@@ -382,6 +382,11 @@ func (host *vmHost) RunSmartContractCall(input *vmcommon.ContractCallInput) (vmO
 	return
 }
 
+// Close closes all internal instances of the vm
+func (host *vmHost) Close() error {
```
bogdan-rosianu, line 386: comment does not match the behaviour

sasurobert: it does not need to close anything, but it the same as for new vm. this is a feat only for compatibility

bogdan-rosianu (approved):

andrei-marinica (approved):

PR #494, opened 2022-01-17
Closer on v1.2

sasurobert:
Closer on v1.2

### DIFF
--- a/arwen/host/arwen.go
+++ b/arwen/host/arwen.go
@@ -387,6 +387,11 @@ func (host *vmHost) RunSmartContractCall(input *vmcommon.ContractCallInput) (vmO
 	return
 }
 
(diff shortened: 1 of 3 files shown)


arwen/host/arwen.go
```diff
@@ -387,6 +387,11 @@ func (host *vmHost) RunSmartContractCall(input *vmcommon.ContractCallInput) (vmO
 	return
 }
 
+// Close closes all internal instances of the vm
```
bogdan-rosianu, line 390: comment does not match the behaviour

sasurobert: there is nothing to close here. but it is the same for all old and new vms

bogdan-rosianu (approved):

andrei-marinica (approved):

PR #506, opened 2022-02-08
Extra TRACE logs (especially related to gas)

andreibancioiu:
Added extra logs on log-level TRACE, mostly do print gas-related information (gas provided, gas locked, gas remaining).

These logs are very useful on local testnets (testnets started with erdpy).


andreibancioiu:
**Question for review: in elrond-go, the `vmOutput` is checked for nil. Should we perform this check here as well, before the `log.Trace` (it seems it is never nil though)?**

sasurobert (approved):

camilbancioiu (approved):

PR #513, opened 2022-02-16
added clean return data endpoint

matei-p:


arwen/contexts/output.go
```diff
@@ -201,6 +201,14 @@ func (context *outputContext) ClearReturnData() {
 	context.outputState.ReturnData = make([][]byte, 0)
 }
 
+func (context *outputContext) RemoveReturnData(index uint32) {
```
sasurobert, line 204: add comment for public function

arwen/elrondapi/elrondei.go
```diff
@@ -3222,24 +3236,70 @@ func v1_4_getReturnDataSize(context unsafe.Pointer, resultID int32) int32 {
...
+	output.ClearReturnData()
+}
+
+//export v1_4_deleteFromReturnData
+func v1_4_deleteFromReturnData(context unsafe.Pointer, resultID int32) {
+	host := arwen.GetVMHost(context)
+	DeleteFromReturnDataWithHost(host, resultID)
+}
+
+func DeleteFromReturnDataWithHost(host arwen.VMHost, resultID int32) {
```
sasurobert, line 3292: comment on public function

arwen/elrondapi/elrondei.go
```diff
@@ -3222,24 +3236,70 @@ func v1_4_getReturnDataSize(context unsafe.Pointer, resultID int32) int32 {
...
+}
 
-	return int32(len(returnData[resultID]))
+//export v1_4_cleanReturnData
+func v1_4_cleanReturnData(context unsafe.Pointer) {
+	host := arwen.GetVMHost(context)
+	CleanReturnDataWithHost(host)
+}
+
+func CleanReturnDataWithHost(host arwen.VMHost) {
```
sasurobert, line 3276: comment on public function

arwen/elrondapi/elrondei.go
```diff
@@ -3222,24 +3236,70 @@ func v1_4_getReturnDataSize(context unsafe.Pointer, resultID int32) int32 {
...
 	if resultID >= int32(len(returnData)) {
-		return 0
+		return nil
 	}
 
-	err := runtime.MemStore(dataOffset, returnData[resultID])
-	if arwen.WithFault(err, context, runtime.ElrondAPIErrorShouldFailExecution()) {
-		return 0
-	}
+	return returnData[resultID]
```
camilbancioiu, line 3267: `resultID` can be negative here. There should be a check that `resultID` is ≥ 0, or at least cast it to an `uint32`. Contracts written in C can pass -4 here, causing a panic.

arwen/contexts/output.go
```diff
@@ -201,6 +201,14 @@ func (context *outputContext) ClearReturnData() {
...
+	returnData := context.outputState.ReturnData
+	if index >= uint32(len(returnData)) {
+		return
+	}
+	context.outputState.ReturnData = append(returnData[:index], returnData[index+1:]...)
```
camilbancioiu, line 209: Would it make sense to call `ClearReturnData` if `index` is 0 and `len(returnData) == 1`? What happens here if we remove the last element of `returnData`? Hopefully, `append` will return an empty slice, and won't return `nil`.

arwen/hosttest/execution_test.go
```diff
@@ -2895,6 +2897,75 @@ func TestExecution_Mocked_Warm_Instances_Same_Contract(t *testing.T) {
...
+						childInput.GasProvided = 1000
+						returnValue := contracts.ExecuteOnDestContextInMockContracts(host, childInput)
+
+						require.Equal(t, int32(0), returnValue)
+						returnData := elrondapi.GetReturnDataWithHostAndTypedArgs(host, 0)
+						require.Equal(t, zero, string(returnData))
+						returnData = elrondapi.GetReturnDataWithHostAndTypedArgs(host, 1)
+						require.Equal(t, one, string(returnData))
+						returnData = elrondapi.GetReturnDataWithHostAndTypedArgs(host, 2)
+						require.Equal(t, two, string(returnData))
```
camilbancioiu, line 2925: Consider adding cases with negative indices.

camilbancioiu (changes requested):

arwen/contexts/output.go
matei-p: Your hopes are true :) A test was just added.

sasurobert (approved):

camilbancioiu (approved):

PR #514, opened 2022-02-16
Token validation fix + tests

matei-p:


andrei-marinica (approved):

arwen/elrondapi/eei_helpers.go
```diff
@@ -61,7 +61,7 @@ func getESDTRoles(data_buffer []byte) int64 {
 	return result
 }
 
-func validateToken(tokenID []byte) int32 {
+func ValidateToken(tokenID []byte) int32 {
```
sasurobert, line 64: add comment to public funciton

dorin-iancu (dismissed):

dorin-iancu (dismissed):

sasurobert (dismissed):

arwen/elrondapi/eei_helpers.go
```diff
@@ -61,7 +61,8 @@ func getESDTRoles(data_buffer []byte) int64 {
 	return result
 }
 
-func validateToken(tokenID []byte) int32 {
+// ValidateToken - validates the token ID
+func ValidateToken(tokenID []byte) int32 {
```
andrei-marinica, line 65: I think we should return bool here and only convert to int32 when returning the value to the contract.

andrei-marinica (changes requested):

andrei-marinica (approved):

camilbancioiu (approved):

PR #523, opened 2022-03-01
set storage refactor

matei-p:


arwen/contexts/storage.go
```diff
@@ -238,88 +238,139 @@ func (context *storageContext) SetStorage(key []byte, value []byte) (arwen.Stora
...
+	if bytes.Equal(oldValue, value) {
+		return context.storageUnchanged(length, usedCache)
 	}
 
-	if !usedCache || !context.flagUseDifferentGasCostForReadingCachedStorage.IsSet() {
-		metering.UseGas(extraKeyLenGas)
+	context.changeStorageUpdate(key, value, storageUpdates)
+
+	var zero []byte
+	if bytes.Equal(oldValue, zero) {
```
andrei-marinica, line 256: I think len(oldValue) == 0 is more lightweight, but not sure

andrei-marinica (approved):

sasurobert (approved):

arwen/contexts/storage.go
sasurobert: +1 -> use len oldValue == 0

PR #538, opened 2022-04-08
fix double use gas

sasurobert:
fix double use gas


dorin-iancu (approved):

sasurobert:
It was resolved in another PR

PR #553, opened 2022-05-12
Master to big float 12 may

sasurobert:


arwen/elrondapi/bigIntOps.go
```diff
@@ -1198,3 +1205,23 @@ func v1_4_bigIntFinishSigned(context unsafe.Pointer, referenceHandle int32) {
 	gasToUse = math.MulUint64(metering.GasSchedule().BaseOperationCost.PersistPerByte, uint64(len(bigInt2cBytes)))
 	metering.UseAndTraceGas(gasToUse)
 }
+
+//export v1_4_bigIntToString
+func v1_4_bigIntToString(context unsafe.Pointer, bigIntHandle int32, destinationHandle int32) {
```
andrei-marinica, line 1210: Let's also add the base here (2..10, 16), should be easy

matei-p (approved):

andrei-marinica (changes requested):

andrei-marinica (approved):

PR #560, opened 2022-05-19
byte storage counting

matei-p:


arwen/contexts/output.go
```diff
@@ -648,6 +650,13 @@ func mergeOutputAccounts(
 	if rightAccount.CodeDeployerAddress != nil {
 		leftAccount.CodeDeployerAddress = rightAccount.CodeDeployerAddress
 	}
+
+	if rightAccount.BytesAddedToStorage > leftAccount.BytesAddedToStorage {
+		leftAccount.BytesAddedToStorage = rightAccount.BytesAddedToStorage
```
andrei-marinica, line 655: This seems inaccurate, unless by BytesAddedToStorage we mean "maximum bytes added to storage".
(Should be ok if left account bytes added to storage include the ones on the right and the counting happened during merge)

matei-p: currently BytesAddedToStorage is not cleared when the new child context is created, to the merge must take this into account

sasurobert (dismissed):

sasurobert (approved):

PR #567, opened 2022-05-27
Delete account endpoint in VM

matei-p:


arwen/host/execution.go
```diff
@@ -137,6 +137,13 @@ func (host *vmHost) checkGasForGetCode(input *vmcommon.ContractCallInput, meteri
 	return nil
 }
 
+// doRunSmartContractDelete delete a contract directly
+func (host *vmHost) doRunSmartContractDelete(input *vmcommon.ContractCallInput) *vmcommon.VMOutput {
+	vmOutput := host.outputContext.GetVMOutput()
```
sasurobert, line 142: add checkPermission here. Do the same in the upgrade function as well. Let's check permission before doing the actual action, not after.

sasurobert (approved):

andrei-marinica (approved):

PR #569, opened 2022-05-29
Create contractCallInput only if gasLimit is positive

lcswillems:
Closes https://github.com/ElrondNetwork/wasm-vm/issues/568


sasurobert:
self.send.direct() is going to be changed to not allow data transfer. Data is user for processing if destination is a SC.

lcswillems:
Great!

PR #580, opened 2022-06-14
fix asyncCallback transfer

sasurobert:


dorin-iancu (approved):

camilbancioiu (approved):

dorin-iancu (changes requested):
Needs an empty result actually.

andrei-marinica (dismissed):

dorin-iancu (dismissed):

arwen/host/asyncCall.go
```diff
@@ -413,6 +417,9 @@ func (host *vmHost) createCallbackContractCallInput(
 	arguments := [][]byte{
 		big.NewInt(int64(destinationVMOutput.ReturnCode)).Bytes(),
 	}
+	if host.flagFixAsyncCallArguments.IsSet() && destinationVMOutput.ReturnCode == vmcommon.Ok {
+		arguments[0] = []byte{0}
```
andrei-marinica, line 421: There's some duplicated code. I think it would help to have a function that converts ReturnCode -> []byte, that also takes the flag into account.

There are 3 places where it would be used, we get rid of many if-s.

andrei-marinica (changes requested):

camilbancioiu (approved):

andrei-marinica (approved):

PR #581, opened 2022-06-15
Transfer value only logs & test

matei-p:


arwen/contexts/validator.go
```diff
@@ -47,6 +47,22 @@ func (validator *wasmValidator) verifyFunctions(instance wasmer.InstanceHandler)
 	return nil
 }
 
+var protectedFunctions = map[string]bool{
+	"internalVMErrors":  true,
+	"transferValueOnly": true}
```
sasurobert, line 52: protect the following as well: 	generalSCRIdentifier = "writeLog"
	signalError          = "signalError"
	completedTxEvent     = "completedTxEvent"

arwen/contexts/validator.go
```diff
@@ -47,6 +47,22 @@ func (validator *wasmValidator) verifyFunctions(instance wasmer.InstanceHandler)
...
+
+func (validator *wasmValidator) verifyProtectedFunctions(instance wasmer.InstanceHandler) error {
+	for functionName := range instance.GetExports() {
```
sasurobert, line 55: this is not backward compatible. please use flag.

sasurobert: use flagManagedCryptoAPI - that is the next one which is going to be activated.

arwen/contexts/validator.go
matei-p: done

arwen/contexts/validator.go
matei-p: done

arwen/contexts/runtime.go
```diff
@@ -684,6 +684,12 @@ func (context *runtimeContext) VerifyContractCode() error {
 			logRuntime.Trace("verify contract code", "error", err)
 			return err
 		}
+	} else {
```
sasurobert, line 687: not with else - be specific. add another if separatly

matei-p: fixed

sasurobert (dismissed):

arwen/contexts/output.go
```diff
@@ -292,6 +297,13 @@ func (context *outputContext) TransferValueOnly(destination []byte, sender []byt
 	senderAcc.BalanceDelta = big.NewInt(0).Sub(senderAcc.BalanceDelta, value)
 	destAcc.BalanceDelta = big.NewInt(0).Add(destAcc.BalanceDelta, value)
 
+	context.WriteLogWithIdentifier(
+		context.host.Runtime().GetContextAddress(),
+		[][]byte{destination, sender, value.Bytes()},
```
andrei-marinica, line 302: Isn't it more intuitive to have the sender, then the destination?

arwen/contexts/output.go
```diff
@@ -292,6 +297,13 @@ func (context *outputContext) TransferValueOnly(destination []byte, sender []byt
 	senderAcc.BalanceDelta = big.NewInt(0).Sub(senderAcc.BalanceDelta, value)
 	destAcc.BalanceDelta = big.NewInt(0).Add(destAcc.BalanceDelta, value)
 
+	context.WriteLogWithIdentifier(
```
sasurobert, line 300: write log only if value.cmp(zero) > 0

arwen/contexts/output.go
matei-p: updated

arwen/contexts/output.go
matei-p: fixed

andrei-marinica (approved):

PR #582, opened 2022-06-20
set to nil

sasurobert:
set to nil


wasmer/instance.go
```diff
@@ -263,7 +265,9 @@ func (instance *Instance) Clean() {
 
 		if instance.Memory != nil {
 			instance.Memory.Destroy()
+			instance.Memory = nil
 		}
+		instance.instance = nil
```
iulianpascalau, line 270: some unit tests would have been great

arwen/contexts/runtime.go
```diff
@@ -104,6 +104,7 @@ func NewRuntimeContext(
 func instanceEvicted(_ interface{}, value interface{}) {
 	localContract, ok := value.(instanceAndMemory)
 	if !ok {
+		value = nil
```
AdoAdoAdo, line 107: this can be removed, as it only sets the local copy to nil

PR #583, opened 2022-06-20
v1_4 enable epochs handler integration

sstanculeanu:


arwen/contexts/storage.go
```diff
@@ -21,36 +20,32 @@ type storageContext struct {
 	stateStack                    [][]byte
 	elrondProtectedKeyPrefix      []byte
 	arwenStorageProtectionEnabled bool
-
-	useDifferentGasCostForReadingCachedStorageEpoch uint32
-	flagUseDifferentGasCostForReadingCachedStorage  atomic.Flag
+	enableEpochsHandler           vmcommon.EnableEpochsHandler
```
camilbancioiu, line 23: Because the `EnableEpochsHandler` is already on the `host`, it should not be a field of the `StorageContext`. The `RuntimeContext` already uses the `EnableEpochsHandler` from the `host` directly.

sstanculeanu: pushed

camilbancioiu (approved):

matei-p (approved):

PR #604, opened 2022-08-31
wasmer import refactor

andrei-marinica:
Decoupled import declaration from wasmer.

Import declaration belongs with the elrondapi part of the VM, while the wasmer imports are a specific executor implementation.

Done in preparation for wasmer 2.2.


arwen/elrondapi/manBufOps.go
```diff
@@ -119,11 +119,6 @@ func ManagedBufferImports(imports *wasmer.Imports) (*wasmer.Imports, error) {
 		return nil, err
 	}
 
-	imports, err = imports.Append("mBufferSetBytes", v1_4_mBufferSetBytes, C.v1_4_mBufferSetBytes)
```
camilbancioiu, line 122: Why was `mBufferSetBytes()` removed?

andrei-marinica: Was duplicate! 
Discovered because I actually added a check for duplicates!

camilbancioiu (dismissed):

arwen/elrondapi/manBufOps.go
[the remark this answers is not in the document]
andrei-marinica: It was duplicate. I discovered that because I added a check for duplicates.

sasurobert (approved):

arwen/host/arwen.go
```diff
@@ -138,43 +137,12 @@ func NewArwenVM(
...
-	if err != nil {
-		return nil, err
-	}
-
-	imports, err = cryptoapi.CryptoImports(imports)
-	if err != nil {
-		return nil, err
-	}
+	imports := elrondapimeta.NewEIFunctions()
+	PopulateAllImports(imports)
```
camilbancioiu, line 142:
```suggestion
	err := PopulateAllImports(imports)
	if err != nil {
		return err
	}
```

camilbancioiu (dismissed):

sasurobert (approved):

PR #608, opened 2022-09-13
Promises node refactor merge from master

andrei-marinica:


andrei-marinica:
Very badly merged, starting from scratch

PR #612, opened 2022-09-20
Wasmer interface refactor, part 2: `executorinterface` package redesign

andrei-marinica:


arwen/contexts/validator.go
```diff
@@ -6,7 +6,7 @@ import (
 
 	vmcommon "github.com/ElrondNetwork/elrond-vm-common"
 	"github.com/ElrondNetwork/wasm-vm/arwen"
-	"github.com/ElrondNetwork/wasm-vm/wasmer"
+	"github.com/ElrondNetwork/wasm-vm/executorinterface"
```
sasurobert, line 9: "executor" keep it simple.

executorinterface/executorInstanceBuilder.go
```diff
@@ -0,0 +1,22 @@
+package executorinterface
+
+type CompilationOptions struct {
```
sasurobert, line 3: add comment

executor/executorInstanceBuilder.go
```diff
@@ -0,0 +1,23 @@
+package executor
```
camilbancioiu, line 1: Consider renaming the file to `instanceBuilder.go`. It's already in the package `executor`.

camilbancioiu (approved):

sasurobert (approved):

executor/executorInstanceBuilder.go
andrei-marinica: I prefixed them all, it can be sometimes hard otherwise to seek for the file in the IDE

camilbancioiu (approved):

sasurobert (approved):

PR #614, opened 2022-09-21
Feat/v1_5 merge from master (via promises-node-refactor)

andrei-marinica:


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/ElrondNetwork/elrond-go-core v1.1.13
 	github.com/ElrondNetwork/elrond-go-crypto v1.0.0
 	github.com/ElrondNetwork/elrond-go-logger v1.0.7
-	github.com/ElrondNetwork/elrond-vm-common v1.3.17-0.20220914092453-aec8d55cd00c
+	github.com/ElrondNetwork/elrond-vm-common v1.3.17-0.20220919101133-149c0edcc821
```
camilbancioiu, line 10: `elrond-vm-common` now has `v1.3.17`, consider referencing it here.

mandos-go/json/write/writeAccount.go
```diff
@@ -41,6 +41,9 @@ func AccountsToOJ(accounts []*mj.Account) oj.OJsonObject {
 		if len(account.Owner.Value) > 0 {
 			acctOJ.Put("owner", bytesFromStringToOJ(account.Owner))
 		}
+		if len(account.DeveloperReward.Original) > 0 {
+			acctOJ.Put("developerRewards", bigIntToOJ(account.DeveloperReward))
```
camilbancioiu, line 45: For another time: would it be possible to refactor this using the generics in Go 1.18?

go.mod
sasurobert: this is with async calls - which has a separate vm-common for now.

mandos-go/json/write/writeAccount.go
andrei-marinica: we're still using go 1.13

camilbancioiu (approved):

PR #626, opened 2022-10-11
Remove static references to wasmer in VM

andrei-marinica:
Also renamed InstanceBuilder -> Executor.

All removed except imports, which are coming in the next PR.


ovstinga (dismissed):

ovstinga (approved):

wasmer/wasmerStatic.go
```diff
@@ -0,0 +1,41 @@
+package wasmer
```
camilbancioiu, line 1: The functions in `wasmerStatic.go` should be private and only accessible via a `wasmerExecutor` instance.

camilbancioiu (approved):

wasmer/wasmerStatic.go
andrei-marinica: Yes, will do in the end

sasurobert (commented):
Nice interfaces!

PR #629, opened 2022-10-14
Executor imports interface, via code generation, part 3

andrei-marinica:


ovstinga (approved):
The interfaces look very good @andrei-marinica . Nice job!

matei-p (approved):

PR #631, opened 2022-10-14
Failed refactor attempt before the big imports refactor

andrei-marinica:


andrei-marinica:
Failed refactor, went down another route in the end

PR #652, opened 2022-11-21
Run smart contract via blockchain hook

matei-p:


arwen/host/execution.go
```diff
@@ -254,6 +254,23 @@ func (host *vmHost) ExecuteOnDestContext(input *vmcommon.ContractCallInput) (vmO
 	scExecutionInput := input
 
 	blockchain := host.Blockchain()
+
+	isSmartContract := blockchain.IsSmartContract(input.RecipientAddr)
```
sasurobert, line 258: I think this code should be more inside exeucteOnDestContextNoBuiltInFunction.

arwen/host/execution.go
```diff
@@ -429,6 +446,14 @@ func (host *vmHost) ExecuteOnSameContext(input *vmcommon.ContractCallInput) erro
 
 	managedTypes, blockchain, metering, output, runtime, _, _ := host.GetContexts()
 
+	vmType, err := vmcommon.ParseVMTypeFromContractAddress(input.RecipientAddr)
+	if err != nil {
+		return err
+	}
+	if !bytes.Equal(runtime.GetVMType(), vmType) {
+		return vmcommon.ErrInvalidVMType
```
sasurobert, line 454: do we really need this ?

arwen/host/execution.go
sasurobert: Wouldn't it be better to have the code deep inside func (host *vmHost) execute(input *vmcommon.ContractCallInput) error ?

arwenmandos/exec.go
```diff
@@ -101,6 +103,17 @@ func (ae *ArwenTestExecutor) GetVM() vmi.VMExecutionHandler {
 
 // GetVMHost returns de vm Context from the vm context map
 func (ae *ArwenTestExecutor) GetVMHost() arwen.VMHost {
+	skipStackLevels := 2
```
sasurobert, line 106: this is some hardcoded stuff. Can't we do better ?

mandos-go/expression/interpreter/interpreter.go
```diff
@@ -181,6 +183,15 @@ func (ei *ExprInterpreter) InterpretString(strRaw string) ([]byte, error) {
 	return ei.interpretNumber(strRaw, 0)
 }
 
+// GetVMType yields the configured VM type, which is used for generating SC addresses.
+// Will yield default value [0, 0] is not explicitly configured.
+func (ei *ExprInterpreter) GetVMType() *[core.VMTypeLen]byte {
+	if ei.VMType == nil {
+		ei.VMType = &[core.VMTypeLen]byte{0, 0}
```
sasurobert, line 190: I think this should yield WASM VM type by default.

sasurobert (changes requested):

arwen/host/execution.go
matei-p: refactored and used a logic similar with the one from builtin functions

arwen/host/execution.go
matei-p: removed it

arwenmandos/exec.go
matei-p: @andrei-marinica

mandos-go/expression/interpreter/interpreter.go
matei-p: @andrei-marinica

andrei-marinica: This component has no access to the VM. I made it editable, but there is no mechanism yet to set it.
There is a `var DefaultVMType = []byte{0xF, 0xF}`, should I rather have that as default?

arwenmandos/exec.go
andrei-marinica: I don't really know what that variable means. This is the mandos executor, but I guess we could extract to a constant.

andrei-marinica (approved):

arwenmandos/exec.go
camilbancioiu: Deleted, this is unnecessary. This code appears to have been copied from the old `GetVMHost()`, when it had VM hook logs. But here, `arwenmandos.GetVMHost()` was not called from within VM hooks, so the whole logging code here did nothing useful.

camilbancioiu (dismissed):

camilbancioiu (approved):

andrei-marinica (approved):

PR #665, opened 2023-01-03
EI generator copied from `feat/wasmer2` and slightly adapted

andrei-marinica:
The point is to have all the code generators also available in master. They were starting to diverge.

Also merged with the changes from the Go Linter.

The wasmer2-related generators are disabled in master.


arwen/elrondapi/generate/eiGenWriteHeaders.go
```diff
@@ -0,0 +1,21 @@
+package elrondapigenerate
+
+import (
+	"fmt"
+)
+
+func autoGeneratedHeader(out *eiGenWriter) {
```
ovstinga, line 7: There are some left over places where we can use this.

ovstinga (approved):

arwen/elrondapi/generate/eiGenWriteHeaders.go
andrei-marinica: Yes, but I'd rather fix this in the next PR

matei-p (approved):

PR #687, opened 2023-01-26
Getting the rebranding branch to master

matei-p:


.github/workflows/build-test.yml
```diff
@@ -5,7 +5,7 @@ on:
     branches:
       - master
   pull_request:
-    branches: [ master, development, rc/* ]
```
andrei-marinica, line 8: I think we should get rid of this, the build CI should run on all branches.

andrei-marinica (approved):

camilbancioiu (approved):

PR #692, opened 2023-01-31
Adapt generator for wasm2 locals cost

ovstinga:
PS: some leftover rebranding in generator, `Draft` until we discuss (**FIXED**)


camilbancioiu (dismissed):

vmhost/vmhooks/generate/eiGenWriteRustWasmerMeteringHelpers.go
```diff
@@ -30,9 +27,24 @@ func WriteRustWasmerMeteringHelpers(out *eiGenWriter) {
 	var content string
 	for fileScanner.Scan() {
 		line := fileScanner.Text()
+		if strings.ToLower(line) == localCostName {
```
andrei-marinica, line 30: This is ok for now, but moving forward, we should create 2 lists: one with wasmer2 operators, and one with whatever else we use in the executor.

andrei-marinica (dismissed):

vmhost/vmhooks/generate/eiGenWriteRustWasmerMeteringHelpers.go
ovstinga: Agreed

andrei-marinica (approved):

vmhost/vmhooks/generate/eiGenWriteRustWasmerMeteringHelpers.go
camilbancioiu: For now, you could extract this into a separate function which returns a boolean, deciding whether to keep the opcode or `continue` over it. Then any future changes will be isolated there.

Indeed, the `if` is a little crude, but it's completely sufficient for now.

camilbancioiu (approved):

PR #713, opened 2023-02-27
Fix test executor switch

camilbancioiu:
This PR fixes the `Makefile` and adds a skipping condition to the tests that require both Wasmer 1 and Wasmer 2 in the same run. This condition only allows the tests to run when `wasmer1` is selected, to avoid conflicts. The tests are just as meaningful as before.

### DIFF
--- a/integrationTests/json/scenariosExecutorLogs_test.go
+++ b/integrationTests/json/scenariosExecutorLogs_test.go
@@ -3,11 +3,15 @@ package vmjsonintegrationtest
 import (
 	"testing"
 
+	"github.com/multiversx/mx-chain-vm-go/testcommon/testexecutor"
 	"github.com/multiversx/mx-chain-vm-go/wasmer"
 	"github.com/multiversx/mx-chain-vm-go/wasmer2"
 )
 
 func TestRustCompareAdderLog(t *testing.T) {
+	if !testexecutor.IsWasmer1Allowed() {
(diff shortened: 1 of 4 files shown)


integrationTests/json/scenariosExecutorLogs_test.go
```diff
@@ -242,6 +290,10 @@ func TestEgldEsdtSwapLog(t *testing.T) {
 }
 
 func TestPingPongEgldLog(t *testing.T) {
+	if !testexecutor.IsWasmer1Allowed() {
```
sasurobert, line 293: why are you skipping these things ?

ovstinga (approved):

sasurobert (approved):

integrationTests/json/scenariosExecutorLogs_test.go
camilbancioiu: Wasmer 2 can conflict with Wasmer 1 during initialization. These tests are intended to initialize both Wasmer 1 and Wasmer 2, which leads to a SIGABRT panic if there was a Wasmer 2 executor already initialized.

The function `IsWasmer1Allowed()` only returns true if the default executor is Wasmer 1, which ensures there is no previous Wasmer 2 executor initialized. This means the tests can be safely executed.

sasurobert (approved):

ovstinga (approved):

PR #727, opened 2023-03-14
fix output transfer

matei-p:


sasurobert (commented):
Can you add a unit test for this ?

sasurobert (approved):

andrei-marinica (approved):

PR #734, opened 2023-05-08
Feat/log events

matei-p:


vmhost/contexts/asyncLocal.go
```diff
@@ -63,6 +63,8 @@ func (context *asyncContext) executeAsyncLocalCall(asyncCall *vmhost.AsyncCall)
 		return vmhost.ErrNilDestinationCallVMOutput
 	}
 
+	context.host.Output().CompleteLogEntriesWithCallType("AsyncCall")
```
sasurobert, line 66: create a logEvent for callBacks as well.

vmhost/hostCore/execution.go
```diff
@@ -339,6 +339,8 @@ func (host *vmHost) ExecuteOnDestContext(input *vmcommon.ContractCallInput) (vmO
 		vmOutput, isChildComplete, err = host.executeOnDestContextNoBuiltinFunction(scExecutionInput)
 	}
 
+	host.Output().CompleteLogEntriesWithCallType("ExecuteOnDestContext")
```
sasurobert, line 342: I do not think this is a good place to add - as you call this endpoint from asyncCall and after transferAndExecute as well.

vmhost/hostCore/execution.go
```diff
@@ -676,6 +680,9 @@ func (host *vmHost) CreateNewContract(input *vmcommon.ContractCreateInput) (newC
 	if err != nil {
 		return
 	}
+
+	host.Output().CompleteLogEntriesWithCallType("DeployFromSource")
```
sasurobert, line 684: Add UpgradeFromSource, UpgradeSmartContract and CreateSmartContract logs as well.

PR #736, opened 2023-05-15
test/contracts:Add mxsc.json and remove wasm files

CostinCarabas:


test/adder/output/adder.mxsc.json
```diff
@@ -0,0 +1,66 @@
+{
+    "buildInfo": {
```
andrei-marinica, line 2: Also delete the no longer used .wasm files. I pointed in the mx-sdk-rs PR how to easily automate this.

andrei-marinica (changes requested):

PR #747, opened 2023-06-16
Removed wrapper math functions

iulianpascalau:
- removed wrapper math functions

### DIFF
--- a/math/overflowHandler.go
+++ b/math/overflowHandler.go
@@ -9,12 +9,18 @@ type overflowHandler struct {
 	err error
 }
 
+// NewOverflowHandler creates a new instance of type overflowHandler
+// This struct is not concurrent safe and should be instantiated for each chain of operations required to be executed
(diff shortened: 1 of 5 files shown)


BeniaminDrasovean (approved):

math/overflowHandler.go
```diff
@@ -0,0 +1,47 @@
...
+	err error
+}
+
+func NewOverflowHandler() *overflowHandler {
+	return &overflowHandler{}
+}
+
+// AddInt64 will add the two provided values. If an overflow occurs, the error will be stored internally
+func (handler *overflowHandler) AddInt64(a, b int64) int64 {
+	s := a + b
```
camilbancioiu, line 18: Should return `builtinMath.MaxInt64` early if `handler.err != nil`. Same for `MulInt64`.

BeniaminDrasovean (approved):

PR #779, opened 2023-09-13
fixed transferAndExecute log for eGLD transfer

sasurobert:


vmhost/contexts/output.go
```diff
@@ -401,12 +397,17 @@ func (context *outputContext) Transfer(
 		context.WriteLogWithIdentifier(
 			sender,
 			[][]byte{value.Bytes(), destination},
-			[][]byte{[]byte(""), input},
+			[][]byte{[]byte("DirectCall"), input},
```
iulianpascalau, line 400: create a const for this and use it also on L435?

camilbancioiu (dismissed):

vmhost/contexts/output.go
sasurobert: Done

andrei-marinica (approved):

camilbancioiu (approved):

iulianpascalau (approved):

PR #837, opened 2024-03-25
Transfer and execute by user

sasurobert:
Implementation of Transfer and execute by user


vmhost/hostCore/host.go
```diff
@@ -537,6 +542,18 @@ func (host *vmHost) AreInSameShard(leftAddress []byte, rightAddress []byte) bool
 	return leftShard == rightShard
 }
 
+// IsAllowedToExecute returns true if the special opcode is allowed to be run by the address
+func (host *vmHost) IsAllowedToExecute(opcode string) bool {
```
laurci, line 546: This is not used anywhere. I think it needs a test or two.

laurci (changes requested):

PR #842, opened 2024-04-13
add mandos for non payable malfunction

BiancaIalangi:


andrei-marinica:
Test already integrated at the time of the Barnard release.

PR #892, opened 2024-10-30
feat async v3

laurci:
refactors, callback order, gas lock, results accumulation


vmhost/hostCore/host.go
```diff
@@ -99,7 +100,7 @@ func NewVMHost(
 	if check.IfNil(hostParameters.EnableEpochsHandler) {
 		return nil, vmhost.ErrNilEnableEpochsHandler
 	}
-	err := core.CheckHandlerCompatibility(hostParameters.EnableEpochsHandler, allFlags)
+	err := core.CheckHandlerCompatibility(hostParameters.EnableEpochsHandler, []core.EnableEpochFlag{})
```
sasurobert, line 103: here you need allFlags

vmhost/contexts/output.go
```diff
@@ -567,22 +567,55 @@ func (context *outputContext) DeployCode(input vmhost.CodeDeployInput) {
...
+	metering := context.host.Metering()
+
+	callType := runtime.GetVMInput().CallType
+
 	runtime.AddError(err, runtime.FunctionName())
 
 	returnCode := context.resolveReturnCodeFromError(err)
 	returnMessage := context.resolveReturnMessageFromError(err)
 
+	if context.host.EnableEpochsHandler().IsFlagEnabled(vmhost.AsyncV3Flag) && callType == vm.AsynchronousCallBack {
```
sasurobert, line 607: I do not think this is ok, also you should check whether the async call was of type v3.

vmhost/hostCore/execution.go
```diff
@@ -1150,6 +1148,14 @@ func (host *vmHost) checkFinalGasAfterExit() error {
 	return nil
 }
 
+func (host *vmHost) checkValidFunctionName(name string) error {
+	if name == "" {
```
sasurobert, line 1152: I think we have a function validator when we deploy the contracts. Use that.

vmhost/vmhooks/baseOps.go
```diff
@@ -1452,6 +1453,144 @@ func CreateAsyncCallWithTypedArgs(host vmhost.VMHost,
 	return 0
 }
 
+// CreateAsyncV3Call VMHooks implementation.
+// @autogenerate(VMHooks)
+func (context *VMHooksImpl) CreateAsyncV3Call(
+	destOffset executor.MemPtr,
```
sasurobert, line 1459: everything should be managed. not PTR. Or have the same call with managed buggers as well.

vmhost/vmhooks/baseOps.go
```diff
@@ -1452,6 +1453,144 @@ func CreateAsyncCallWithTypedArgs(host vmhost.VMHost,
...
+	dataOffset executor.MemPtr,
+	dataLength executor.MemLength,
+	successOffset executor.MemPtr,
+	successLength executor.MemLength,
+	errorOffset executor.MemPtr,
+	errorLength executor.MemLength,
+	gas int64,
+	extraGasForCallback int64,
+) int32 {
+	calledSCAddress, err := context.MemLoad(destOffset, vmhost.AddressLen)
```
sasurobert, line 1498: gas should be used before making any kind of load.

vmhost/asyncCall.go
```diff
@@ -23,21 +23,28 @@ type AsyncCall struct {
 	CallbackClosure []byte
 
 	IsBuiltinFunctionCall bool
+	IsAsyncV3             bool
```
sasurobert, line 26: this is not backward compatible.

PR #894, opened 2024-11-21
mandos test update

andrei-marinica:


test/features/composability/forwarder/output/forwarder-dbg.mxsc.json
```diff
@@ -0,0 +1,1523 @@
+{
```
andrei-marinica, line 1: `*dbg.mxsc.json` is here by accident, it's not used in tests
call a sc-meta all clean before running the test copy tool

BiancaIalangi (approved):

mihaicalinluca (approved):

PR #896, opened 2025-01-14
mandos updates for framework v0.55 (master)

andrei-marinica:


andrei-marinica:
Mandos tests already updated in Barnard: https://github.com/multiversx/mx-chain-vm-go/pull/940

PR #897, opened 2025-01-16
mask internal libs errors

laurci:


vmhost/vmhooks/cryptoei.go
```diff
@@ -69,6 +71,12 @@ func (context *VMHooksImpl) Sha256(
 
 	result, err := crypto.Sha256(data)
 	if err != nil {
+
+		if enableEpochsHandler.IsFlagEnabled(vmhost.MaskInternalDependenciesErrorsFlag) {
+			context.FailExecution(vmhost.ErrSha256Hash)
```
sasurobert, line 76: I think it is shorter to have:

err = vmhost.ErrSha256Hash inside the if, and context fail and return remain the same.

sasurobert: This suggestion can be used everywhere.

laurci: I simplified using assignment everywhere

sasurobert (approved):

andreiblt1304 (approved):

PR #906, opened 2025-02-18
update scenario v1.5.0

BiancaIalangi:


andrei-marinica:
mx-scenario-go already at 1.6.0

PR #907, opened 2025-02-24
transferAndExecute and ExecuteOnDest with error return. new endpoints…

sasurobert:
…. for complex usage

## Related PRs:
- https://github.com/multiversx/mx-vm-executor-rs/pull/54


laurci (dismissed):

vmhost/vmhooks/managedei.go
```diff
@@ -1189,6 +1247,44 @@ func (context *VMHooksImpl) ManagedMultiTransferESDTNFTExecute(
...
+	}
+
+	return TransferESDTNFTExecuteWithTypedArgsWithFailure(
+		host,
+		vmInput.destination,
+		transfers,
+		gasLimit,
+		[]byte(vmInput.function),
+		vmInput.arguments,
+		true,
```
ccorcoveanu, line 1284: shouldn't this be false?

andrei-marinica (dismissed):

vmhost/vmhooks/managedei.go
sasurobert: no. this is the only place we want it to be true. this is the new function.

ccorcoveanu: only asked because the main function is called `ManagedMultiTransferESDTNFTExecuteWithReturn` (WithReturn, not Fail) 👍

laurci: @ccorcoveanu good catch. it should be false. The tests call the inner `TransferESDTNFTExecuteWithTypedArgsWithFailure` directly and didn't catch this. Will add more tests.

ccorcoveanu (approved):

vmhost/vmhooks/managedei.go
laurci: added more integration tests. resolving this now

laurci (approved):

ccorcoveanu (approved):

PR #934, opened 2025-06-18
mandos - forwarder reject payment callback call value

andrei-marinica:


andrei-marinica:
Test already added in the Barnard release.

PR #953, opened 2025-08-04
feat: Add balance invariants check

sasurobert:
This change adds a new balance check to the VM host to ensure that all token transfers are handled correctly. The check is performed at the end of each transaction.

The following checks are implemented:
- Conservation of base currency (EGLD). (Currently disabled due to incomplete VMOutput)
- No pending back transfers for synchronous calls. (Currently disabled due to incomplete VMOutput)

The check for ESDT token conservation is not yet implemented as it requires more information about the data structures used to store ESDT balances in the VMOutput.

The base currency and back transfers checks are also disabled for now because the VMOutput from the vmhost does not contain the full picture of the balance changes (e.g., sender's balance change for the initial transfer and gas fee).

### DIFF
--- a/vmhost/balances.go
+++ b/vmhost/balances.go
@@ -1,0 +1,96 @@
+package vmhost
+
+import (
+	"math/big"
+	"strings"
+
+	"github.com/multiversx/mx-chain-core-go/data/vm"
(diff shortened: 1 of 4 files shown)


sasurobert:
moved to another PR.

PR #954, opened 2025-08-05
feat: Add unsafe execution mode

sasurobert:
I've introduced an unsafe execution mode for smart contracts. When activated, certain errors that would normally halt the execution of a smart contract will be ignored. This allows you to build more resilient contracts that can handle unexpected errors without crashing.

The following changes were made:

- I added `activateUnsafeMode` and `deactivateUnsafeMode` hooks to allow smart contracts to control the unsafe mode.
- I added a `FailExecutionConditionally` function that only fails the execution if the unsafe mode is not active.
- I replaced `FailExecution` with `FailExecutionConditionally` in various hooks where it makes sense to allow for non-fatal errors.
- I added mock tests to verify the new functionality.

### DIFF
--- a/executor/vmHooks.go
+++ b/executor/vmHooks.go
@@ -16,6 +16,7 @@ type VMHooks interface {
 	ManagedMapVMHooks
 	SmallIntVMHooks
 	CryptoVMHooks
+	UnsafeVMHooks
 }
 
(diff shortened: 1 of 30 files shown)


sasurobert:
added to another PR.

PR #957, opened 2025-08-06
Add unit tests for vmhost

sasurobert:


laurci (dismissed):

laurci (approved):

vmhost/contexts/asyncParams_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"testing"
+
+	"github.com/multiversx/mx-chain-core-go/data/vm"
+	vmcommon "github.com/multiversx/mx-chain-vm-common-go"
+	"github.com/multiversx/mx-chain-vm-go/mock/mockery"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAddAsyncArgumentsToOutputTransfers(t *testing.T) {
+	t.Parallel()
```
andrei-marinica, line 15: Does it help for these tests to run in parallel? It looks like they are only testing object construction.

andrei-marinica (approved):

vmhost/contexts/asyncParams_test.go
sasurobert: always help. it is faster.

PR #978, opened 2026-01-27
Integrate agg changes on crypto

ssd04:


crypto/signing/bls/bls.go
```diff
@@ -46,5 +46,25 @@ func (b *bls) VerifySignatureShare(publicKey []byte, message []byte, sig []byte)
 
 // VerifyAggregatedSig verifies aggregated signature of BLS MultiSig
 func (b *bls) VerifyAggregatedSig(pubKeysSigners [][]byte, message []byte, aggSig []byte) error {
-	return b.multiSigner.VerifyAggregatedSig(pubKeysSigners, message, aggSig)
+	pubKeys, err := b.convertBytesToPubKeys(pubKeysSigners)
```
ssd04, line 49: this should behave as before, until now this convert was done in crypto multi signer

ssd04: we can analyse to cache here pubkeys and avoid conversion (which takes quite some time ~60ms)

go.mod
```diff
@@ -9,7 +9,7 @@ require (
 	github.com/gogo/protobuf v1.3.2
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/multiversx/mx-chain-core-go v1.4.0
-	github.com/multiversx/mx-chain-crypto-go v1.3.0
+	github.com/multiversx/mx-chain-crypto-go v1.3.1-0.20260127130843-3c6c4667127e
```
raduchis, line 12: do not forget actual release

crypto/signing/bls/bls.go
raduchis: yes... it  will add some overhead here... the 60 ms...

raduchis (approved):

ssd04:
This will be on hold for now. Changed https://github.com/multiversx/mx-chain-crypto-go/pull/53 to add new methods for the signing methods.

